PR #4, opened 2023-01-17
Define each connection with BaseConnection

rgildein:
- moved all connections from manager.py to network.py
- defined BaseConnection abstract class
- defined BaseSubprocessConnection for SshPortForwardSubprocess and SshuttleSubprocess


esunar (approved):
Base structure is solid. However we may need further enhancements while creating connection objects.

PR #6, opened 2023-01-17
add controller connect w/ exponential retries

rgildein:
- move defining `self.connections[controller_config.name]` before controller connect
- add retry function for controller w/ exponential retries
- fix empty controller_name and controller_uuid when connection without JUJU_DATA

usage example: https://pastebin.ubuntu.com/p/SpPGbz2gvg/

fixes: #7


juju_spell/connections/manager.py
```diff
@@ -121,6 +126,57 @@ class Connection:
...
+    password: str,
+    cacert: str,
+):
+    """Direct connection to controller without JUJU_DATA.
+
+    This is a helper function for connecting to a controller with simple exponential
+    retry and with fix for missing controller_name and controller_uuid.
+    """
+    start = time.time()
+    wait: float = 1.5
```
esunar, line 144: This should come from config
global config->local config->controller config->default

juju_spell/connections/manager.py
```diff
@@ -1,18 +1,23 @@
...
 from juju import juju
+from juju.errors import JujuConnectionError
 
 from juju_spell.config import Controller
 from juju_spell.settings import JUJUSPELL_DEFAULT_PORT_RANGE
 
 logger = logging.getLogger(__name__)
 
 MAX_FRAME_SIZE = 6**24
+CONENCTION_TIMEOUT = 60  # seconds
```
esunar, line 20: This should come from config
global config->local config->controller config->default

esunar (changes requested):

juju_spell/connections/manager.py
```diff
@@ -121,6 +126,57 @@ class Connection:
...
+            break
+        except JujuConnectionError:
+            # Note(rgildein): Connection will raise JujuConnectionError if endpoint
+            # is unreachable. This can happen, for example, when port forwarding is
+            # through a subprocess and the process has not yet started.
+            logger.info("%s connection to controller %s failed", uuid, name)
+            await asyncio.sleep(wait)
+            if time.time() - start >= CONENCTION_TIMEOUT:
+                raise
+            wait = 1.5 ** (attempt - 2)  # exponential wait 1.5^(x-2)
```
esunar, line 167: It should be constant

esunar (changes requested):

juju_spell/connections/manager.py
rgildein: Right now I moved everything to `settings.py` and I reported an issue #21, so that we can implement it as we discussed.

esunar (approved):

PR #10, opened 2023-01-18
fix(juju_spell/settings.py): Use SNAP_REAL_NAME to get the real use's home directory

jneo8:


esunar (approved):

rgildein (approved):
LGTM, I tested it and it work fine.

PR #12, opened 2023-01-18
rename username to user & enable empty model in model_mapping

rgildein:
- rename username to user in config
- enable empty model name for model_mapping in config


jneo8 (changes requested):
Do we also need to update the cli arguement from username to user?

rgildein:
> Do we also need to update the cli arguement from username to user?

I was also thinking about it, but I think it make sense only when someone is copy-pasting from juju configs to juju-spell config.

esunar (approved):

esunar (approved):

PR #13, opened 2023-01-18
remove VersionCMD

rgildein:
After discussion, we came to the conclusion that the version should be displayed only when `--version` is used and the `version' command should display the version of the controller and not the juju-spell.

I'll add `VersionCMD` back in next PR, but it will return version of controller.


jneo8 (changes requested):
Maybe a short unit test for `_run_dispatcher`

rgildein:
> Maybe a short unit test for `_run_dispatcher`

I created #15 issue to add 100% coverage, but you are right I should not be lazy to create `test_cmd.py` and add test for `_run_dispatcher`. :+1:

rgildein:
Example of usage: https://pastebin.ubuntu.com/p/tjYn7dBDW6/

During making this changes I hit an issue ([lp#1998783](https://bugs.launchpad.net/juju/+bug/1998783)) in Juju.

esunar (approved):

PR #32, opened 2023-01-31
set up environemnt for functional tests

rgildein:


tests/functional/setup_lxd.py
```diff
@@ -0,0 +1,150 @@
...
+            juju_path,
+            "bootstrap",
+            "--verbose",
+            f"--config='{config.name}'",
+            cloud_name,
+            new_controller_name,
+        ]
+        process = subprocess.check_output(cmd)
+    try:
+        process.wait(timeout=60 * 5)
```
esunar, line 83: can you put timeout values to constants

tests/functional/setup_lxd.py
```diff
@@ -0,0 +1,150 @@
...
+from juju.client._definitions import CloudsResult
+from juju.controller import Controller
+
+logger = logging.getLogger(__name__)
+
+
+def _check_juju_deps() -> str:
+    """Check if Juju is installed."""
+    try:
+        return subprocess.check_output(["which", "juju"]).decode().strip()
```
esunar, line 38: should we install juju if not installed?

PR #34, opened 2023-02-02
prepare LXD environemnt (client + 2 controllers)

rgildein:
- creates "client" LXD container with JujuSpell installed
- bootstrap N controllers

TODO:
- [x] figure out why `setup_environment` works when each step is run individually, but not at once
- [x] figure out why `execute` raises [this](https://pastebin.ubuntu.com/p/2QcTSWCBSd/) errors (command is executed, but this error is always there)
- [x] creates config file for JujuSpell and copy paste it to the client
- [x] add simple test with ping comand

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -1,0 +1,120 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2023 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""JujuSpell configuration for functional tests."""
+import logging
+import os
+import subprocess
+from pathlib import Path
+from typing import Callable, Generator
+
+import pylxd.models
+import pytest
+from pylxd import Client
+
+from tests.functional.setup_lxd import cleanup_environment, setup_environment
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_SERIES = "jammy"
+CLIENT_CONAINER = pytest.StashKey[str]()
+CONTROLLERS_CONTAINERS = pytest.StashKey[list]()
+
+
+def check_dependencies(*dependencies) -> None:
+    """Check if dependencies are installed."""
+    errors = []
+    for dependence in dependencies:
+        try:
+            subprocess.check_output(["which", dependence])
+        except subprocess.CalledProcessError:
+            errors.append(f"Missing {dependence} dependency.")
+
+    if errors:
+        raise RuntimeError(os.linesep.join(errors))
+
+
+def build_snap(build: bool) -> Path:
+    """Build JujuSpell snap."""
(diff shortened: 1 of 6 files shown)


rgildein:
Not included in this PR:

- [ ] README.md for functional tests with explanation
- [ ] configure GitHub workflow

tests/functional/conftest.py
```diff
@@ -0,0 +1,120 @@
...
+
+DEFAULT_SERIES = "jammy"
+CLIENT_CONAINER = pytest.StashKey[str]()
+CONTROLLERS_CONTAINERS = pytest.StashKey[list]()
+
+
+def check_dependencies(*dependencies) -> None:
+    """Check if dependencies are installed."""
+    errors = []
+    for dependence in dependencies:
```
esunar, line 40: dependence -> dependency

tests/functional/setup_lxd.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def cleanup_environment():
+    """Clean up LXD environment.
+
+    Remove all instances with names starting with CONTAINER_PREFIX.
+    """
+    client = Client()
+    for instance in client.instances.all():
+        if instance.name.startswith(CONTAINER_PREFIX):
```
esunar, line 208: We should create a unique string for each run

esunar (changes requested):

tests/functional/conftest.py
rgildein: I think that bout are valid expressions. Do we really need to switch here?

rgildein:
I added README.md and configured GH workflows in #52.

Examples of usage with `--no-build` and `--keep-env`.

tests/functional/conftest.py
esunar: just because everyone uses dependency like dependency injection

tests/functional/setup_lxd.py
```diff
@@ -0,0 +1,214 @@
...
+    """Clean up LXD environment.
+
+    Remove all instances with names starting with CONTAINER_PREFIX.
+    """
+    client = Client()
+    for instance in client.instances.all():
+        if str(session_uuid) in instance.name:
+            if keep_env:
+                if (
+                    instance.status_code != 102
```
esunar, line 206: can you define a constant for 102

tests/functional/test_connection.py
```diff
@@ -0,0 +1,23 @@
+import json
+
+
+def test_help(juju_spell_run):
+    """Test printing --help."""
+    help_flag = juju_spell_run("--help")
+    assert help_flag.exit_code == 0
+
+    help_default = juju_spell_run()
+    assert help_default.exit_code == 127
```
esunar, line 10: can you define a constant for 127

tests/functional/setup_lxd.py
```diff
@@ -0,0 +1,214 @@
...
+    print(f"LXD: {container.name} is running")
+    return container
+
+
+def create_client_instance(
+    session_uuid: str, series: str, snap_path: Union[str, Path]
+) -> Instance:
+    """Create client instance."""
+    client = create_container(session_uuid, "client", series)
+    sleep(5)  # TODO: we need to wait for container to be fully readyx
```
esunar, line 90: Random sleeps!! Can we make it a function to run a simple command on container until it succeeds?

esunar (changes requested):

esunar:
Can you also update Makefile

tests/functional/conftest.py
jneo8: dependence is abstract whereas dependency is more concrete

tests/functional/setup_lxd.py
rgildein: I put TODO there and forgot to do anything about it. I'll create function.

esunar (approved):

PR #35, opened 2023-02-02
Returning Result from each command

rgildein:
- fix returning task result and not list of tasks results (only one task was run)
- add Result object for each task
- refactor add_user format_output function based on Result

### DIFF
--- a/juju_spell/commands/add_user.py
+++ b/juju_spell/commands/add_user.py
@@ -19,7 +19,6 @@ class AddUserCommand(BaseJujuCommand):
         )
 
         return {
-            "uuid": kwargs["controller_config"].uuid,
             "user": user.username,
             "display_name": user.display_name,
             "password": password,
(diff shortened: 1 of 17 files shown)


juju_spell/commands/add_user.py
```diff
@@ -19,7 +19,6 @@ async def execute(self, controller: Controller, **kwargs):
         )
 
         return {
-            "uuid": kwargs["controller_config"].uuid,
```
esunar, line 22: why we have moved this to cli?

esunar (changes requested):
UUID is one of the main outputs of add_user so this should be in command part

juju_spell/commands/add_user.py
rgildein: I moved it because it's not output from add-user juju command, so if you want to use this in composed command you will not need to have uuid of controller, since you have whole controller. I seem logical to me, but I have no strong opinion here.

PR #38, opened 2023-02-06
Add helper script to generate config

peppepetra:


rgildein (changes requested):
Can we make "helpers" folder ignored by snap.

rgildein (approved):
LGTM

PR #39, opened 2023-02-07
fix(juju_spell/filter.py): Fix filter not working problem

jneo8:
The major reason why filter is broken is because the type of filter expression is string not list.
Create a serialize function to fix it.


juju_spell/filter.py
```diff
@@ -18,6 +18,13 @@ def make_controllers_filter(filter_expression):
     and b in [v4,v5,v6].
     """
 
+    def serialize(values):
+        if isinstance(values, str):
+            return set(values.split(","))
+        if isinstance(values, list):
+            return set(values)
+        return values
+
```
rgildein, line 27: not-blocking: we are not testing this function

rgildein (approved):
LGTM

juju_spell/filter.py
jneo8: It's a internal helper function, which should already been tested when run filter unit test. If in the future we make it too complex to understand, I will move it to another place and add unit test to it. But now it's not needed.

PR #44, opened 2023-02-09
Update packages with tests

esunar:


docs/update_packages-retval.json
```diff
@@ -0,0 +1,28 @@
+[
```
rgildein, line 1: same ^

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+@dataclasses.dataclass
+class Container:
+    updates: list[Update]
+
+
+class UpdatePackages(BaseJujuCommand):
+    async def execute(self, controller: Controller, **kwargs):
+        default_model = kwargs["controller_config"].model_mapping["default"]
```
rgildein, line 44: Why we want to run this only on default model? Why not using `--models` or all models?

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,142 @@
...
+    def get_apps_to_update(self, juju_status, updates):
+        apps_to_update = []
+        for update in updates:
+            for app, app_status in juju_status.applications.items():
+                if re.match(update["application"], app):
+                    apps_to_update.append((update, app, list(app_status.units.keys())))
+        return apps_to_update
+
+    async def get_juju_status(self, controller, default_model):
+        juju_status_map = await StatusCommand.execute(
```
rgildein, line 132: I think it would be simpler to run:
```python
model = controller.get_model(<model_name>)
model.connect()
status = model.get_status()
model.disconnect()
```

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,142 @@
...
+        app_list = []
+        for app in update["packages-to-update"]:
+            app_list.append(app["app"])
+        package_list = " ".join(app_list)
+        return package_list
+
+    def get_apps_to_update(self, juju_status, updates):
+        apps_to_update = []
+        for update in updates:
+            for app, app_status in juju_status.applications.items():
```
rgildein, line 126: Can we use `model.applications` and than `model.application[<app>].units` to get units, instead of getting it from status?

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,142 @@
...
+
+@dataclasses.dataclass
+class Container:
+    updates: list[Update]
+
+
+class UpdatePackages(BaseJujuCommand):
+    async def execute(self, controller: Controller, **kwargs):
+        default_model = kwargs["controller_config"].model_mapping["default"]
+        updates = self.get_patch_config(kwargs)
```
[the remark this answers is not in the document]
esunar: fix

juju_spell/commands/update_packages.py
esunar: fixed

juju_spell/commands/update_packages.py
esunar: will be done in future

docs/update_packages-input.json
```diff
@@ -0,0 +1,57 @@
+{
```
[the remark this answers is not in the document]
esunar: they should be as is but needs to be referenced

docs/update_packages-retval.json
esunar: they should be as is but needs to be referenced

juju_spell/cli/update_packages.py
```diff
@@ -0,0 +1,131 @@
...
+        raise JujuSpellError(f"permission denied to read config file {path}") from error
+
+
+def get_patch_config(file_path: str):
+    patch = load_patch_file(file_path)
+    applications: List[Application] = []
+    for app in patch["applications"]:
+        packages_to_update: List[PackageToUpdate] = []
+        for package in app.get("packages_to_update", []):
+            p = PackageToUpdate(
```
rgildein, line 117: nitpick: can we rename it from `p` -> `package_to_update`

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,179 @@
...
+        self, controller: Controller, models: Optional[List[str]] = None, **kwargs
+    ) -> Optional[Result]:
+        """Run update."""
+        return await self.make_updates(controller=controller, models=models, **kwargs)
+
+    async def dry_run(
+        self, controller: Controller, models: Optional[List[str]] = None, **kwargs
+    ) -> Optional[Result]:
+        """Run update with --dry-run flag."""
+        return await self.make_updates(controller=controller, models=models, **kwargs)
```
rgildein, line 75: I think you missed `dry_run=True` here.

rgildein (changes requested):
One nitpick and one change request comment. Do we want to add unit tests here? Can we provide example of usage? Thanks

juju_spell/commands/update_packages.py
rgildein: In that case can we report it as issue? So we will not forgot about it. Thanks

docs/update_packages-input.json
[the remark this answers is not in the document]
rgildein: I think we should put them at least in sub-folder, something like 'commands/update-packages'. When we create our spec we can reference to that.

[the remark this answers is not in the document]
esunar: /docs/commands/update-packages or /commands/update-packages ?

juju_spell/commands/update_packages.py
esunar: This is implemented

juju_spell/commands/update_packages.py
esunar: it is in kwargs if I add it will create duplicate variable error

docs/update_packages-input.json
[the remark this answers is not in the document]
rgildein: `docs/commands/update-packages` I think

juju_spell/cli/update_packages.py
esunar: done

juju_spell/commands/update_packages.py
rgildein: I see, Do you think that we should pop it from kwargs in assignment or here? It's not obvious that here and I think that we should run `self.make_updates` fir `dry_run=True` argument.

esunar: can you check again

docs/update_packages-input.json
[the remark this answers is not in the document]
esunar: done

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,191 @@
...
+            updates=kwargs["patch"],
+            dry_run=True,
+        )
+
+    async def make_updates(
+        self,
+        controller: Controller,
+        models: Optional[List[str]],
+        dry_run: bool,
+        model_mapping: Dict[str, List[str]],
```
rgildein, line 94: The `model_mapping` can be `None` as we found out recently if I'm not wrong?

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,191 @@
...
+            if line.startswith("Inst"):
+                _, name, from_version, to_version, *others = line.split(" ")
+
+            # Unpacking software-properties-common (0.99.9.11) over (0.99.9.10)
+            elif line.startswith("Unpacking"):
+                _, name, from_version, _, to_version, *others = line.split(" ")
+
+            tv = to_version.strip("()[]")
+            fv = from_version.strip("()[]")
+            n = name.strip(" ")
```
rgildein, line 147: nitpick: I really think we should avoid using variables like this. Here there are used in next line, but if not it's hard to track what is going on, when I see `tv` variable. It's tv show? :D

rgildein (changes requested):
One question (potentially change request), one nitpick comment and I think you need to remove empty file `generate-config.py`.

juju_spell/commands/update_packages.py
esunar: it can be empty cannot be none

juju_spell/commands/update_packages.py
esunar: done

esunar:
Also empty file removed

juju_spell/commands/update_packages.py
rgildein: Empty means None, or not?

scripts/generate-config.py
```diff
@@ -1,112 +0,0 @@
-#!/usr/bin/env python3
```
rgildein, line 1: Now I'm confused, last time I saw this as empty file and now it's not.

rgildein (commented):
We are still missing unit tests for `juju_spell/commands/update_packages.py`, but otherwise LGTM.

jneo8 (commented):
Leave 1 suggestion

juju_spell/commands/update_packages.py
esunar: []

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,193 @@
...
+                    PackageUpdateResult(
+                        package=name, from_version=from_version, to_version=to_version
+                    )
+                )
+
+        return packages
+
+    def get_update_command(self, app: Application, dry_run: bool):
+        """Generate command according to flags."""
+        template = UPDATE_TEMPLATE + ("--dry-run" if dry_run else "")
```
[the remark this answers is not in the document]
esunar: added tests

scripts/generate-config.py
esunar: fixed

juju_spell/commands/update_packages.py
```diff
@@ -134,13 +134,15 @@ def parse_result(self, result: str):
...
+            if line.startswith("Inst "):
                 _, name, from_version, to_version, *others = line.split(" ")
 
             # Unpacking software-properties-common (0.99.9.11) over (0.99.9.10)
             elif line.startswith("Unpacking"):
-                _, name, from_version, _, to_version, *others = line.split(" ")
+                _, name, to_version, _, from_version, *others = line.split(" ")
 
             to_version = to_version.strip("()[]")
             from_version = from_version.strip("()[]")
```
rgildein, line 148: not-blocking: I think we can move this to separate function.

juju_spell/commands/update_packages.py
rgildein: I'm confused, because in this [line](https://github.com/canonical/juju-spell/blob/2833ae4686034cb31d6c66aaaa0d53cd8e3dae55/juju_spell/commands/base.py#L139) we are checking if `model_mapping` is `None`, but we are using `model_mapping: Dict[str, List[str]]` everywhere.

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,195 @@
...
+    def set_apps_to_update(self, model: Model, updates: Updates, dry_run: bool):
+        """Set units to applications.
+
+        Finds the matching applications and set the units of these applications as a
+        List[UpdateResult] to application.
+        """
+        for update in updates.applications:
+            command = self.get_update_command(app=update, dry_run=dry_run)
+            for app, app_status in model.applications.items():
+                if re.match(update.name_expr, app):
```
rgildein, line 178: What will happens if in patch file I use application, which does not exists? It will be quietly ignored?

I think that we can create issue for this and for check that unit exists (check my previous comment).

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,195 @@
...
+
+    async def run_updates_on_model(self, model: Model, updates: Updates):
+        """Run updates on model.
+
+        Runs the command on unit and parses the result and assigns it to each unit.
+        """
+        for app in updates.applications:
+            for result in app.results:
+                for unit in result.units:
+                    juju_unit = model.units[unit.unit]
```
rgildein, line 118: If unit does not exist, the KeyError will be raised and that will not provide good information to user (hard to debug).

I think that we can report an issue for this and handle it in separate PR.

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,195 @@
...
+            if line.startswith("Inst "):
+                _, name, from_version, to_version, *others = line.split(" ")
+
+            # Unpacking software-properties-common (0.99.9.11) over (0.99.9.10)
+            elif line.startswith("Unpacking"):
+                _, name, to_version, _, from_version, *others = line.split(" ")
+
+            to_version = to_version.strip("()[]")
+            from_version = from_version.strip("()[]")
+            name = name.strip(" ")
```
rgildein, line 149: I think we can move this to the separate function.

juju_spell/commands/update_packages.py
```diff
@@ -0,0 +1,195 @@
...
+                    update.results.append(
+                        UpdateResult(application=app, units=unit_updates)
+                    )
+
+    def set_success_flags(
+        self, unit: UnitUpdateResult, expected: List[PackageToUpdate]
+    ):
+        """Set success flag for each unit."""
+        expected_set = set([(e.version, e.package) for e in expected])
+        real_set = [(p.to_version, p.package) for p in unit.packages]
```
rgildein, line 192: nitpick: I think we should avoid using single character variable name, even in list comprehension.

rgildein, line 192: You missed one `set` and you do not need to use list comprehension here. `set` will handle the generator
```suggestion
        expected_set = set((e.version, e.package) for e in expected)
        real_set = set((p.to_version, p.package) for p in unit.packages)
```

tests/unit/cli/test_cli_update_packages.py
```diff
@@ -0,0 +1,77 @@
...
+    file_path = tmp_path / f"{uuid.uuid4()}.yaml"
+    with open(file_path, "w", encoding="utf8") as file:
+        file.write(input_yaml)
+
+    result = load_patch_file(file_path)
+    assert result == yaml.safe_load(io.StringIO(input_yaml))
+
+
+@pytest.mark.parametrize("input_yaml", [TEST_PATCH])
+def test_get_patch_config(tmp_path, input_yaml):
```
rgildein, line 70: Function `load_patch_file` should be mocked here, so if that function failed only one test will failed.

tests/unit/cli/test_cli_update_packages.py
```diff
@@ -0,0 +1,77 @@
+import argparse
```
rgildein, line 1: not-blocking: The name of test file do not need to be named `test_cli_*`, since we are using `cli` sub-folder.

tests/unit/commands/test_update_packages.py
```diff
@@ -0,0 +1,344 @@
...
+"""
+
+UNIT_UPDATE_COMMAND = (
+    "sudo apt-get update ; sudo apt-get"
+    " --option=Dpkg::Options::=--force-confold"
+    " --option=Dpkg::Options::=--force-confdef dist-upgrade --upgrade -y  "
+)
+
+
+@pytest.fixture
```
rgildein, line 123: not-blocking: It's common practice to use `conftest.py` for defining any fixtures.

rgildein (changes requested):

rgildein:
Can you also provide description for this PR + some example how did you tested it. Thanks

juju_spell/commands/update_packages.py
esunar: I do not think it is a problem, it is a feature. Because I want to update all prometheus applications I do not care if exits or not. In results it will show updated packages as empty.

juju_spell/commands/update_packages.py
esunar: We get unit names from juju in set_apps_to_update. There is not such option.

juju_spell/commands/update_packages.py
esunar: I agree but in limited scope like comprehensions it does no harm

tests/unit/cli/test_cli_update_packages.py
esunar: The other files were like these to I follow the structure

tests/unit/commands/test_update_packages.py
esunar: Bu these fixtures are specific to this file.

PR #55, opened 2023-02-14
model_mapping improvements

esunar:
Changed controllers.model_mapping to Dict[str, List[str]] thus we will be able to add more than one model as default

get_filtered_models now handles the model_mapping

### DIFF
--- a/juju_spell/commands/base.py
+++ b/juju_spell/commands/base.py
@@ -125,9 +125,6 @@ def _apply_model_mappings(
     Returns:
         list of models replaced with values from model_mappings
     """
-    if model_mappings is None or len(model_mappings) <= 0:
-        return models
-
     results = []
     for model in models:
         results.extend(model_mappings.get(model, [model]))
--- a/juju_spell/commands/base.py
+++ b/juju_spell/commands/base.py
@@ -44,7 +44,7 @@ class BaseJujuCommand(metaclass=ABCMeta):
     @staticmethod
     async def get_filtered_models(
         controller: Controller,
-        model_mappings: Dict,
+        model_mappings: Dict[str, List[str]],
         models: Optional[List[str]] = None,
     ) -> AsyncGenerator[Tuple[str, Model], None]:
         """Get filtered models for controller.
--- a/juju_spell/commands/base.py
+++ b/juju_spell/commands/base.py
@@ -50,11 +50,13 @@ class BaseJujuCommand(metaclass=ABCMeta):
         """Get filtered models for controller.
 
         If models is None, then all models for controller will be returned.
+        If the model_mapping[model] exits for specific model it will be replaced by the
+        list of values from model_mapping[model] from config.
         """
         if models is None or len(models) <= 0:
             all_models = await controller.list_models()
         else:
-            all_models = _filter_models(models, model_mappings)
+            all_models = _apply_model_mappings(models, model_mappings)
 
         for model_name in all_models:
             model = await controller.get_model(model_name)
@@ -109,7 +111,7 @@ class BaseJujuCommand(metaclass=ABCMeta):
         ...
 
 
-def _filter_models(
+def _apply_model_mappings(
     models: List[str], model_mappings: Dict[str, List[str]]
 ) -> List[str]:
     """Replace the models with values from model_mappings.
--- a/juju_spell/commands/base.py
+++ b/juju_spell/commands/base.py
@@ -18,7 +18,7 @@
 import dataclasses
 import logging
 from abc import ABCMeta, abstractmethod
-from typing import Any, AsyncGenerator, List, Optional, Tuple
+from typing import Any, AsyncGenerator, Dict, List, Optional, Tuple
 
 from juju.controller import Controller
 from juju.model import Model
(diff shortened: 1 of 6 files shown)


juju_spell/commands/base.py
```diff
@@ -43,15 +43,18 @@ def __init__(self):
...
         If models is None, then all models for controller will be returned.
         """
         all_models = await controller.get_models()
+        extended_models = _get_extended_models(models, model_mappings)
         for model_name in all_models:
-            if not models or model_name in models:
+            if not extended_models or model_name in extended_models:
                 model = await controller.get_model(model_name)
                 yield model_name, model
                 await model.disconnect()
```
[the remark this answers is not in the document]
rgildein: Next suggestion, after seeing we already supports the list of models in model_mapping.

```python
    @staticmethod
    async def _filter_models(
        controller: Controller,
        model_mappings: Dict[str, List[str]],
        models: Optional[List[str]] = None,
    ) -> List[str]:
        """Return filtered list of models names."""
        if models is None:
            return await controller.list_models()

        filtered_models = []
        for model in models:
            filtered_models.extend(model_mappings.get(model, [model]))

        return filtered_models

    @staticmethod
    async def get_filtered_models(
        controller: Controller,
        model_mappings: Dict[str, List[str]],
        models: Optional[List[str]] = None,
    ) -> AsyncGenerator[Tuple[str, Model], None]:
        """Get filtered models for controller.

        If models is None, then all models for controller will be returned.
        """
        filtered_models = await _filter_models(controller, model_mappings, models)
        
        for model_name in filtered_models:
            model = await controller.get_model(model_name)
            yield model_name, model
            await model.disconnect()
```

rgildein (changes requested):

juju_spell/commands/base.py
```diff
@@ -102,3 +105,23 @@ async def execute(
...
+    for model in models:
+        if items := model_mappings.get(model):
+            results.extend(items)
+        else:
+            results.append(model)
+
+    if len(results) > 0:
+        return results
+
+    return None
```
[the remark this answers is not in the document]
esunar: added docstring

juju_spell/commands/base.py
```diff
@@ -43,18 +43,23 @@ def __init__(self):
 
     @staticmethod
     async def get_filtered_models(
-        controller: Controller, models: Optional[List[str]] = None
+        controller: Controller,
+        model_mappings: Dict[str, List[str]],
```
jneo8, line 47: **suggestion:** I will say it's hard for me to first time understand what is model_mapping and what is models.

I think a docstring here to provide more details will be really helpful.

juju_spell/commands/base.py
```diff
@@ -102,3 +107,27 @@ async def execute(
                 will contain the config for the selected controller
         """
         ...
+
+
+def _filter_models(
```
jneo8, line 112: **nitpick:** 

The naming of the function and what it really does is not match.
Maybe we need to know it's more about model mapping?

juju_spell/commands/status.py
```diff
@@ -14,7 +14,9 @@ async def execute(
     ) -> Dict[str, FullStatus]:
         """Get status for selected models in controller."""
         output = {}
-        async for name, model in self.get_filtered_models(controller, models):
+        async for name, model in self.get_filtered_models(
+            controller, models, kwargs["controller_config"].model_mapping
```
jneo8, line 18: **issue:** 
What if model_mapping doesn't exists?
We don't provide the default for model_mapping so it will raise KeyError

juju_spell/commands/base.py
```diff
@@ -102,3 +107,27 @@ async def execute(
...
+    models: List[str], model_mappings: Dict[str, List[str]]
```
jneo8, line 113: **issue:** 

The type annotation of model_mapping is not correct.
Should be `Optional[Dict[str, List[str]]]`

jneo8 (changes requested):

juju_spell/commands/status.py
rgildein: The `model_mapping` is required in config, so it will be there.

juju_spell/commands/base.py
rgildein: Since `model_mapping` is required I do not think this is needed.

juju_spell/commands/base.py
```diff
@@ -102,3 +107,27 @@ async def execute(
...
+    If --model parameter is provided searches the map for matching model, if found
+    puts the corresponding values from controller.model_mapping into results if not
+    found puts the model itself to results.
+    Args:
+        models: list of models from input
+        model_mappings: mappings from config file
+    Returns:
+        list of models replaced with values from model_mappings
+    """
+    if model_mappings is None or len(model_mappings) <= 0:
```
rgildein, line 126: Since we have right now two required keys (`lma` and `default`) in `model_mapping` it's not needed to check len.

rgildein (commented):
Can you add unit tests for `_filter_models` function? I think it's useful to tests also function like this.

juju_spell/commands/base.py
esunar: removed

rgildein (approved):
LGTM

PR #57, opened 2023-02-15
feat(juju_spell/config.py): Create default configuration options

jneo8:


esunar (approved):
I seems ok

esunar (changes requested):
controller1: 30
controller2: 20
controller3: 43

personal.controller > global.controller > personel.default > global.default

personal.config
default:
  timeout: 20
controller1
  timeout: 30
controller2
controller3


global.config
default:
  timeout: 21
controller1
  timeout: 24
controller2
controller3
  timeout: 43

jneo8:
> I could not see the most extensive `codeconstant < global.default.config < global.config < local.default.config < local.config ` test.

We have tests for *apply_default*, *merge_config* and *load_config*.
The logic been hidden behind those three function.
It's possible if we don't patch any them to create this test. Do you think we need this one?

esunar (approved):

juju_spell/config.py
```diff
@@ -199,28 +209,78 @@ class Config:
...
+    template: confuse.MappingTemplate,
+) -> Config:
+    """Return valid config if source match template, else raise ConfigError."""
     try:
         _config = RootView([source])
-        valid_config = _config.get(JUJUSPELL_CONFIG_TEMPLATE)
-        logger.info("config was validated")
-        config = Config(**valid_config)
-        return config
+        valid_config = _config.get(template)
```
rgildein, line 219: nitpick: can we add debug message here? It will be useful.

juju_spell/config.py
```diff
@@ -199,28 +209,78 @@ class Config:
...
+      - name: cbd
+        ...
+    ```
+    The values in default.controller will be apply to every elements in controllers
+    if the value is not exists.
+    """
+    validate_source_match_template(source, JUJUSPELL_DEFAULT_CONFIG_TEMPLATE)
+
+    for key, value in source.get(DEFAULT_KEY, {}).items():
+        target_key = key + "s"
```
rgildein, line 255: Do we really want to use something like this? what if target key is not list? Why not using "controllers" directly?

juju_spell/config.py
```diff
@@ -199,28 +209,78 @@ class Config:
...
+        if isinstance(source.get(target_key), List):
+            source[target_key] = [
+                {**value, **target_value} for target_value in source[target_key]
```
rgildein, line 258: Will this work properly? If we define
```yaml
default:
  controller: 
    connection:
      timeout: 5
```
it will override whole connection in each controller.

rgildein (changes requested):
I think that this will not work properly, can you add unit test to simulate
```yaml
default:
  controller:
    connection:
      timeout: 30
controllers:
- name: 1
  ...
- name: 2
  ...
  connection:
    destination: 1.2.3.4
    timeout: 20
- name: 3
  ...
  connection:
    destination: 1.2.3.4
```
30s is expected timeout for c1 anc c3, for c2 it should be 20. However I think c3 will lost connection.destination.

juju_spell/config.py
jneo8: I am not sure what kind of debug message you want. It will raise error if not validated. It's either error or no output. I don't think this is needed.

juju_spell/config.py
jneo8: Ye, this is a bug. I need to figure out how to fix it.

juju_spell/config.py
rgildein: Where and how? I think that ring now our config file has strict format, so why adding strange magic, which is working only for controllers now.
If any new target come, we will need to update our template and if that target is not list we will need to create new rule how it will be updated, so I do not see any reason why creating it like this now.

rgildein (changes requested):
I tried this with simple config:
```yaml
default:
  controller:
    user: rgildein
controllers:
- uuid: ...
  user: admin
- uuid: ...
  # no user defined
```
And I got an error 
```bash
configuration file validation failed with error: default.controller must be a list, not dict 
```

Can you test the functionality and provide example of your `config.yaml` and `personal.yaml`.
Thanks

jneo8:
> I tried this with simple config:
> 
> ```yaml
> default:
>   controller:
>     user: rgildein
> controllers:
> - uuid: ...
>   user: admin
> - uuid: ...
>   # no user defined
> ```
> 
> And I got an error
> 
> ```shell
> configuration file validation failed with error: default.controller must be a list, not dict 
> ```
> 
> Can you test the functionality and provide example of your `config.yaml` and `personal.yaml`. Thanks

Fixed. The example is already in doc.

PR #58, opened 2023-02-15
Refactor/connection config

jneo8:
Move connection from top layer info controller's config.

The config format is really confuse people after we have the default config. We shouldn't allow two different way to provide the default.

The PR is following PR after #57

### DIFF
--- a/juju_spell/config.py
+++ b/juju_spell/config.py
@@ -77,7 +77,7 @@ class PortRange(String):
     """A template specific for port range."""
 
     def convert(self, value: Any, view: confuse.ConfigView) -> range:
-        """Check and convert port-range string to tuple."""
+        """Check and convert port_range string to tuple."""
         value = super().convert(value, view)
         start_port, end_port = value.split(":")
         return range(int(start_port), int(end_port))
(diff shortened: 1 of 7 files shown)


esunar (approved):

juju_spell/config.py
```diff
@@ -137,6 +137,13 @@ def value(self, view, template=None):
                             String(DESTINATION_REGEX, "Invalid jump definition")
                         )
                     ),
+                    "port_range": confuse.Optional(
```
rgildein, line 140: this should be `port-range`

rgildein (changes requested):
Small fix required.

juju_spell/config.py
jneo8: There are two reasons I change all `port-range` to `port_range`.

1. Same format for all the config. We use variables with underscore like `ca_cert`, `display_name`, etc. But we only use dash for `port-range`.
2. Because we use confuse to load yaml and load it into python's datacalss. Confuse seems not covert dash to underscore for us, which should be a basic behavior for python's configuration library, unfortunately confuse is not in the role. So to have a *clear* definition in python's dataclass, we need to change it to underscore.

esunar: I agree with @jneo8 to compatible with python standards it would be good to be underscore.

rgildein: My bad, I thought that in code you used `port-range`, but you renamed it everywhere.
Yes we need to use `_` for those, because we are passing it to dataclass object, where we could not have `port-range` as variable.

rgildein (approved):
LGTM

PR #61, opened 2023-02-16
Fix get_filtered_models parameter order.

esunar:


rgildein (approved):
LGTM, I tested it and it's working fine (we need to add functional tests for this)

PR #62, opened 2023-02-16
Fix the documentation with respect to config.

esunar:


docs/configuration.md
```diff
@@ -51,7 +51,7 @@ controllers:
 
 ## Default config
 
-The key inside *default* will provide the default value to `{key}s` if the value is not exists.
+The key inside *default* will provide the default value to `controller` if the value is not exists.
```
rgildein, line 54: This is not true, we are still using `key + s`.

rgildein (changes requested):

rgildein (approved):

PR #67, opened 2023-02-21
dropped duplicated requirements

rgildein:
We do not need to have requirements in two places.


esunar (approved):

esunar:
@rgildein can we apply this to setup.cfg and requirements.txt?

PR #80, opened 2023-02-28
Add list-models sub command to list models from controllers

chanchiwai-ray:
This pull request add new sub-command to list all models from a controllers or all controllers. It includes the following features:

- [x] User can use `--refresh` to get latest models information directly from the controllers
- [x] User can specify `--filter` to list models from specific controllers.
- [ ] Support local caching for models information when running without `--refresh` (ToDo)
- [ ] Add tests

[This is a WIP]


juju_spell/commands/list_models.py
```diff
@@ -0,0 +1,22 @@
...
+    async def execute(
+        self, controller: Controller, **kwargs
+    ) -> Dict[str, Union[List, bool]]:
+        """List models from the local cache or from the controllers."""
+        outputs = {"refresh": False, "models": []}
+        if kwargs["refresh"]:
+            outputs["refresh"] = True
+            outputs["models"] = await controller.list_models()
+        else:
+            outputs["models"] = ["TODO: read from cache; not implemented."]
```
esunar, line 18: you need to also change [this function](https://github.com/canonical/juju-spell/blob/main/juju_spell/commands/base.py#LL47C25-L47C25)

esunar (changes requested):

juju_spell/commands/list_models.py
chanchiwai-ray: It's a bit strange to me. That function in the end was used in `status` and `update-packages` commands, both of them don't have `--fresh`. If the changes should also be in that function. I think a more appropriate way is to chain the commands together. i.e move `get_filtered_models` to `list-model` commands. And then `status` command will be like: `list-models` -> `get model status`. What do you think ?

chanchiwai-ray: Okay, I split the `get_filtered_models` into two independent functions that one gets the models names (`get_filtered_model_names`) and the other one gets the models (`get_filtered_models`). And use the `get_filtered_model_names` to actually list the models. The `status` and `update-packages` commands should not be affected by this (i.e. will not get models from the cache), since they don't have `--refresh`, so they always fetch model names from the controllers. If these commands will need this caching feature, we will need to add `--refresh` to them.

juju_spell/commands/base.py
```diff
@@ -44,22 +44,48 @@ def __init__(self) -> None:
...
+        config. By default, models information is fetched from the controller;
+        if refresh is False, then the models information is loaded from local
+        cache.
         """
         if models is None or len(models) <= 0:
-            all_models = await controller.list_models()
+            if refresh:
+                all_models = await controller.list_models()
+            else:
+                all_models = ["TODO: read from cache; not implemented."]
```
jneo8, line 66: **suggestion:** 

Maybe a lru_cache is enough?

https://docs.python.org/3/library/functools.html#functools.lru_cache

And we can use `cache_clear` to force refresh.

juju_spell/commands/base.py
```diff
@@ -44,22 +44,48 @@ def __init__(self) -> None:
...
         model_mappings: Dict[str, List[str]],
         models: Optional[List[str]] = None,
-    ) -> AsyncGenerator[Tuple[str, Model], None]:
-        """Get filtered models for controller.
-
-        If models is None, then all models for controller will be returned.
-        If the model_mapping[model] exits for specific model it will be replaced by the
-        list of values from model_mapping[model] from config.
+        refresh: Optional[bool] = True,
+    ) -> List[str]:
```
jneo8, line 52: **praise:** 

After checking the source code in python-libjuju.
The return should be list, nice one.

jneo8 (approved):
One not blocking suggestion to implement cache.

juju_spell/commands/base.py
chanchiwai-ray: We need to keep it in disk not memory so that it's available during different invocation of command.

jneo8 (approved):

juju_spell/commands/base.py
rgildein: I like idea to splitting this into two functions, but I think that `refresh` should not be an argument here, at-least if we do not want to add `--refresh` to each command. If not this function should always getting models from file and if that file does not exits creating it. The `list-modes` command can recreate this file if `--refresh` was provided,

juju_spell/commands/base.py
```diff
@@ -44,22 +44,48 @@ def __init__(self) -> None:
...
+    ) -> AsyncGenerator[Tuple[str, Model], None]:
+        """Get filtered models for controller.
 
+        If models is None, then all models for controller will be returned.  If
+        the model_mapping[model] exits for specific model it will be replaced
+        by the list of values from model_mapping[model] from config. By
+        default, models information is fetched from the controller; if refresh
+        is False, then the models information is loaded from local cache.
+        """
+        all_models = await BaseJujuCommand.get_filtered_model_names(
```
rgildein, line 86: `self.get_filtered_model_names`

juju_spell/commands/list_models.py
```diff
@@ -0,0 +1,28 @@
...
+            "refresh": kwargs["refresh"],
+            "models": list(
+                await self.get_filtered_model_names(
+                    controller=controller,
+                    models=None,
+                    refresh=kwargs["refresh"],
+                    model_mappings=kwargs["controller_config"].model_mapping,
+                )
+            ),
+        }
```
rgildein, line 26:
```suggestion
        if kwargs["refresh"]:
            reload_models_yaml_file()
        outputs = {
            "refresh": kwargs["refresh"],
            "models": list(
                await self.get_filtered_model_names(
                    controller=controller,
                    models=None,
                    model_mappings=kwargs["controller_config"].model_mapping,
                )
            ),
        }
```

rgildein (changes requested):

juju_spell/commands/base.py
esunar: agree

juju_spell/commands/base.py
```diff
@@ -44,22 +44,48 @@ def __init__(self) -> None:
...
+        returned.  If the model_mapping[model] exits for specific model it will
+        be replaced by the list of values from model_mapping[model] from
+        config. By default, models information is fetched from the controller;
+        if refresh is False, then the models information is loaded from local
+        cache.
         """
         if models is None or len(models) <= 0:
-            all_models = await controller.list_models()
+            if refresh:
+                all_models = await controller.list_models()
```
esunar, line 64: we should find a better to show cached model list without connecting to controller

esunar (changes requested):

juju_spell/commands/base.py
chanchiwai-ray: ACK. We can't if we want to keep it `staticmethod`, but I think we don't need to make it `staticmethod` in the first place which is originated from `get_filtered_models`

juju_spell/commands/list_models.py
chanchiwai-ray: Thanks, and you're right. I think that's better as well.

juju_spell/commands/base.py
chanchiwai-ray: I think there's no better way if we are not going to change codes other than `cli` and `commands`.

juju_spell/commands/base.py
rgildein: Why we cant if it's static method?

chanchiwai-ray: The method itself doesn't need / have `self`

rgildein: It does not, but you can and I think that you should call it like this `all_models = await self.get_filtered_model_names`.
I'm not think that it's an issue.

chanchiwai-ray:
```python


class Foo:
    @staticmethod
    def greeting():
        print("Hello there")

    @staticmethod
    def print_name(name):
        self.greeting()
        print("My name is ", name)

Foo.print_name("Foo")

Traceback (most recent call last):
  File "/tmp/a.py", line 11, in <module>
    Foo.print_name("Foo")
  File "/tmp/a.py", line 8, in print_name
    self.greeting()
NameError: name 'self' is not defined
```
Maybe I am mistaken, I think this will happen

rgildein: No I did not check that you were also calling it in staticmethod, so it makes sense now. My mistake.
In that case I think we should make in class function, since it's calling another class function. (You already did it, so :+1:)

juju_spell/commands/list_models.py
```diff
@@ -0,0 +1,29 @@
...
+        outputs = {
+            "refresh": kwargs["refresh"],
+            "models": list(
+                await self.get_filtered_model_names(
+                    controller=controller,
+                    models=None,
+                    model_mappings=kwargs["controller_config"].model_mapping,
+                )
+            )
+            if kwargs["refresh"]
```
rgildein, line 25: not-blocking: I think it would be more readable if we do 
```python
if kwargs["refresh"]:
    models = ...
else:
    models = ...

outputs = {"refresh": kwargs["refresh"], "models": models}
```
rather than hiding in in dict creation

rgildein (approved):
Overall LGTM, we will add cache functionality in next PR

juju_spell/commands/list_models.py
chanchiwai-ray: Updated

esunar (approved):

Issue #83, opened 2023-03-01
Structure discussion


esunar:

jneo8:
A Composability way to compose atomic operation.
It should be something similar to FP's *railway programming*.

For example, a function level interface always return a Result Object.
The result object can tell us whether the operation is success or failing.

The result should be return/continue to next step. Do we need to have a confirm from user.

esunar:
**First**: I would like to have an ability to request confirmation from the user when executing action. For example when upgrading packages I would like to be able confirm for each unit.

**Second** in future will have a set of structures like `(input:command:output)`
thus if we have commands S,W,X,Y,Z such as
```
S  :X:a
W :W:b
X  a:X:b
Y  b:Y:c
Z  b:Z:a
```
means that we can create composite commands like:
```
S -> X -> Y
S -> X -> Z
W -> Y -> X
.
.
.
```
Our design should allow this.

**Third**: It should be easy to define parameters and commands. If I would like to add new command it should be just adding a file or a bunch of files. The rest of the application should stay as is.

jneo8:
One design idea:

Because the total targets to access are fixed.
`Controller`, `Model`, `Unit`, `Application` and `Machine`. All of them are under a single controller's connection.

So if we can define **operation unit(ops)** which have execute level mapping to each running target.

For example:

```
ops-1 -> ops-2 -> ops-3
```

- *ops-1* works on controller level.
- *ops-2* works on model level.
- *ops-3* works on unit level.

We can define list of operations in module cli. And assignment will for loop in each ops until one of them return error.

The **add-user** command below will become like:

```sh
juju spell add-user --user abc --acl superuser --models m1,m2,m3
```

- ops-1: Controller Level, Create user on controller
- ops-2: Controller Level, Grant user permission on controller
- ops-3: Model Level, Grant model permission on models

During each iteration(ops-1 -> ops-2), is the break point to check if result is error or need confirm.

So the implementation of **upgrade-package** will become something like:

```
- ops-1(update package-a, unit level)
- ops-2(upgrade package-b, unit level)

...
```
A list of operations units, each one represent to upgrade one package on every units.

To do this, we need to:

- An **assignment** package which iterate every level or targets
  - Break point in each operation units.
- An *ops* package which provide operation action on specific level of target.
  - Should only include smallest operation action.
  - Should not have action like: Iterate over models on controller and do something.
- An cli package will need to struct the list of operations units, and inject the operations units to assignment run function.
[...]

PR #86, opened 2023-03-02
Added juju-spell config.

esunar:
* moved load_yaml_file to utils package
* added unit tests for juju-spell config

note: documentation will be added


tests/unit/cli/test_cli_config.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_fill_parser():
+    """Test add additional CLI arguments with ConfigCMD."""
+    parser = mock.MagicMock(spec=argparse.ArgumentParser)
+
+    cmd = ConfigCMD(None)
+    cmd.fill_parser(parser)
+
+    # This one is to check the basic arguments is been added.
+    assert parser.add_argument.call_count == 7
```
chanchiwai-ray, line 36: Just curious, is it really necessary to check these here? Adding more basic arguments in `cli.base` will cause error in here which is not really related to `test_cli_config.py`

tests/unit/cli/test_cli_config.py
```diff
@@ -0,0 +1,65 @@
...
+            mock.call(
+                "--config-set",
+                nargs="*",
+                help="key=value pairs to update application config.",
+                required=False,
+                action=KeyValue,
+            ),
+        ]
+    )
+
```
chanchiwai-ray, line 50: Since you implemented a custom `KeyValue` action, I think it's better to include test case for this action. For example, test if `parser.parse_args(['--config-set a=1 b=2 c=cpu=4])` gives you `{"a": 1, "b":2 , "c": "cpu=4"}`

juju_spell/cli/config.py
```diff
@@ -0,0 +1,137 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""JujuSpell juju status command."""
```
chanchiwai-ray, line 17: `"""JujuSpell juju status command."""`: maybe copy-and-paste error ?

juju_spell/cli/config.py
```diff
@@ -0,0 +1,137 @@
...
+
+from juju_spell import utils
+from juju_spell.cli.base import JujuWriteCMD
+from juju_spell.commands.config import ApplicationConfig, ConfigCommand
+
+logger = logging.getLogger()
+
+
+class ConfigCMD(JujuWriteCMD):
+    """JujuSpell juju status command."""
```
chanchiwai-ray, line 33: Same here

juju_spell/commands/config.py
```diff
@@ -0,0 +1,77 @@
...
+import logging
+from typing import Any, Dict, List
+
+from juju.application import Application
+from juju.controller import Controller
+from juju.model import Model
+
+from juju_spell.commands.base import BaseJujuCommand
+
+logger = logging.getLogger("eeee")
```
chanchiwai-ray, line 12: is it intentional ? `logging.getLogger("eeee")` 🤣

jneo8: I believe is cat.

juju_spell/commands/config.py
```diff
@@ -0,0 +1,77 @@
...
+    config: Dict[str, str]
+
+
+class ConfigCommand(BaseJujuCommand):
+    """Command to configure juju applications."""
+
+    async def execute(self, controller: Controller, *args: Any, **kwargs: Any) -> Dict[str, Any]:
+        """Configure juju applications."""
+        output: Dict[str, Any] = {}
+        async for name, model in self.get_filtered_models(
```
jneo8, line 29: **thought:** 

Not blocking command.

I am thinking about the comment I gave [here](https://github.com/canonical/juju-spell/issues/83#issuecomment-1451355000)

Iterate over models/application/machine inside command really make thing more complex and hard to understand the working flow.

Just want to point out this is the situation I want to avoid.

juju_spell/utils.py
```diff
@@ -96,3 +103,21 @@ def merge_list_of_dict_by_key(key: str, lists: List[List[Dict]]) -> List:
 def random_password(length: int = 30) -> str:
     """Generate random password."""
     return secrets.token_urlsafe(length)
+
+
+def load_yaml_file(path: str) -> Any:
```
jneo8, line 108: **suggestion:** 

Not blocking.
The code here is similar to `load_config_file` function inside `config.py`.
Consider to have a small refactor here.

tests/unit/cli/test_cli_config.py
jneo8: Rather than check add_argument count. I will suggest to actually check the real context give to add_argument.

jneo8 (changes requested):

tests/unit/cli/test_cli_config.py
esunar: extended checks.
remove call_count check

tests/unit/cli/test_cli_config.py
esunar: added tests

juju_spell/cli/config.py
esunar: fixed doc

juju_spell/cli/config.py
esunar: fixed doc

juju_spell/commands/config.py
esunar: killed the cat :)

tests/unit/cli/test_cli_config.py
rgildein: It's not important to check which CLI options from BaseCLI, but it's important to check  that there were used.

juju_spell/cli/config.py
```diff
@@ -0,0 +1,140 @@
...
+    yaml_config = utils.load_yaml_file(path)
+    result: List[ApplicationConfig] = []
+    for app, value_dict in yaml_config.items():
+        application_config = ApplicationConfig(application=app, config=value_dict)
+        result.append(application_config)
+    return result
+
+
+# pylint: disable=too-few-public-methods,signature-differs
+# mypy: allow-untyped-defs
```
rgildein, line 131: We should rather add typing that disabling mypy and pylint.

juju_spell/commands/config.py
```diff
@@ -0,0 +1,77 @@
...
+        juju_app: Application = model.applications.get(update.application)
+        if not juju_app:
+            continue
+
+        await juju_app.set_config(update.config)
+        config.update({update.application: await juju_app.get_config()})
+    return config
+
+
+async def apply_config(app: Application, properties: Dict[str, str], single_property: str) -> dict:
```
rgildein, line 68: I do not understand what is this function trying to do. can you add description in docstring with some examples?

juju_spell/utils.py
rgildein: I agree we should combine that together. And  having one in utils make sense.

rgildein (changes requested):

juju_spell/cli/config.py
esunar: It is impossible to add typing 
```
    def __call__(
        self, parser: ArgumentParser, namespace: Namespace, values: str | Sequence[Any] | None, option_string: str | None = ...
    ) -> None: ...
```
python 3.8 does not support `|`

juju_spell/commands/config.py
esunar: refactored to remove ambigutiy

juju_spell/utils.py
esunar: combined

juju_spell/cli/config.py
rgildein: `from typing import Optional`

esunar: mypy does not like it either

rgildein: That's strange it's should be other issue.

jneo8 (approved):

juju_spell/cli/config.py
rgildein: Can you share the output from mypy.

rgildein:
```python
class KeyValue(argparse.Action):
    """Parse parameters as key=value pairs."""

    def __call__(
        self,
        _: argparse.ArgumentParser,
        namespace: argparse.Namespace,
        values: Union[str, Sequence[Any], None],
        option_string: Optional[str] = None,
    ) -> None:
        setattr(namespace, self.dest, {})
        if values:
            for value in values:
                key, value = value.split("=", 1)
                getattr(namespace, self.dest)[key] = value
```
And from this it's obvious that you could not assume that values is None or List of strings, it cloud be also only string.

chanchiwai-ray (approved):

juju_spell/cli/config.py
esunar: applied your recommendation result:
`juju_spell/cli/config.py:134:0: R0903: Too few public methods (1/2) (too-few-public-methods)`

tests/unit/cli/test_cli_config.py
esunar: extended checks

juju_spell/cli/config.py
esunar: fixed the issue by bypassing too-few-public-methods for python3.8

juju_spell/cli/config.py
```diff
@@ -0,0 +1,149 @@
...
+    """Parse application config to model."""
+    yaml_config = utils.load_yaml_file(Path(path))
+    result: List[ApplicationConfig] = []
+    for app, value_dict in yaml_config.items():
+        application_config = ApplicationConfig(application=app, config=value_dict)
+        result.append(application_config)
+    return result
+
+
+# pylint: disable=too-few-public-methods
```
rgildein, line 134: this is not needed

juju_spell/utils.py
```diff
@@ -15,9 +15,17 @@
...
 from collections import defaultdict
-from typing import Dict, Iterable, List
+from pathlib import Path
+from typing import Any, Dict, Iterable, List
+
+import yaml
+
+from juju_spell.exceptions import JujuSpellError
+
+logger = logging.getLogger()
```
rgildein, line 28: It's better to use new logger, which will inherit from root rather than using root logger.
```suggestion
logger = logging.getLogger(__name__)
```

tests/unit/cli/test_cli_config.py
```diff
@@ -0,0 +1,101 @@
...
+
+def test_add_mutually_exclusive_group_parameters() -> None:
```
rgildein, line 52: not-blocking: define return value for tests function is not needed `-> None`

rgildein (changes requested):
2 small things, but otherwise LGTM

juju_spell/cli/config.py
rgildein: Now I can see that you are hitting still this issue, wired that for me w/ Python3.10 is working.

rgildein: Can you share version of pylint and argparse you are using?

rgildein: I tried it on Python3.8 and lint is not working due the fact that pylxd is not supported, but with some tweak I got same error from pylint. Strange, I do not understand why. Resolving this.

rgildein (approved):
LGTM

PR #87, opened 2023-03-03
feat(juju_spell/commands/add_user.py): Grant user permission when acl…

jneo8:
- Call grant cmd in add-user cmd if `acl` is in the params.
- Grant cmd now will grant model permission for user.

### DIFF
--- a/juju_spell/commands/add_user.py
+++ b/juju_spell/commands/add_user.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, Union
 from juju.controller import Controller
 
 from juju_spell.commands.base import BaseJujuCommand, Result
+from juju_spell.commands.enable_user import EnableUserCommand
 from juju_spell.commands.grant import GrantCommand
 from juju_spell.utils import random_password
 
@@ -26,11 +27,16 @@ class AddUserCommand(BaseJujuCommand):
             display_name=kwargs["display_name"],
         )
 
+        enable_cmd = EnableUserCommand()
+        enable_cmd_result = await enable_cmd.run(controller=controller, **kwargs)
+        if not enable_cmd_result.success:
+            return enable_cmd_result
+
         if kwargs.get("acl"):
             grant_cmd = GrantCommand()
-            grant_result: Result = await grant_cmd.run(controller=controller, **kwargs)
-            if not grant_result.success:
-                return grant_result
+            grant_cmd_result: Result = await grant_cmd.run(controller=controller, **kwargs)
+            if not grant_cmd_result.success:
+                return grant_cmd_result
 
         return {
             "user": user.username,
--- a/juju_spell/commands/add_user.py
+++ b/juju_spell/commands/add_user.py
@@ -1,9 +1,10 @@
 """Command to add users."""
-from typing import Any, Dict
+from typing import Any, Dict, Union
 
 from juju.controller import Controller
 
-from juju_spell.commands.base import BaseJujuCommand
+from juju_spell.commands.base import BaseJujuCommand, Result
+from juju_spell.commands.grant import GrantCommand
 from juju_spell.utils import random_password
 
 __all__ = ["AddUserCommand"]
@@ -12,7 +13,9 @@ __all__ = ["AddUserCommand"]
 class AddUserCommand(BaseJujuCommand):
     """Add user command."""
 
-    async def execute(self, controller: Controller, *args: Any, **kwargs: Any) -> Dict[str, str]:
+    async def execute(
+        self, controller: Controller, *args: Any, **kwargs: Any
+    ) -> Union[Dict[str, str], Result]:
         password = kwargs["password"]
         if len(password) == 0:
             password = random_password()
@@ -22,6 +25,12 @@ class AddUserCommand(BaseJujuCommand):
             password=password,
             display_name=kwargs["display_name"],
         )
+
+        if kwargs.get("acl"):
+            grant_cmd = GrantCommand()
+            grant_result: Result = await grant_cmd.run(controller=controller, **kwargs)
+            if not grant_result.success:
+                return grant_result
 
         return {
             "user": user.username,
--- a/juju_spell/commands/add_user.py
+++ b/juju_spell/commands/add_user.py
@@ -2,7 +2,6 @@
 from typing import Any, Dict, Union
 
 from juju.controller import Controller
-from juju.errors import JujuError
 
 from juju_spell.commands.base import BaseJujuCommand, Result
 from juju_spell.commands.enable_user import EnableUserCommand
@@ -22,18 +21,17 @@ class AddUserCommand(BaseJujuCommand):
         if len(password) == 0:
             password = random_password()
 
-        try:
+        user = await controller.get_user(username=kwargs["user"])
+        if user is None:  # User exists
             user = await controller.add_user(
                 username=kwargs["user"],
                 password=password,
                 display_name=kwargs["display_name"],
             )
-        except JujuError as err:
-            if overwrite:
-                user = await controller.get_user(username=kwargs["user"])
-                await user.set_password(password)  # Reset user's password
-            else:
-                raise err
+            self.logger.info("Create user %s", kwargs["user"])
+        if overwrite:
+            await user.set_password(password)  # Reset user's password
+            self.logger.info("Reset user password")
 
         enable_cmd = EnableUserCommand()
         enable_cmd_result = await enable_cmd.run(
@@ -43,16 +41,12 @@ class AddUserCommand(BaseJujuCommand):
             return enable_cmd_result
 
         if kwargs.get("acl"):
-            try:
-                grant_cmd = GrantCommand()
-                grant_cmd_result: Result = await grant_cmd.run(
-                    controller=controller, overwrite=overwrite, **kwargs
-                )
-                if not grant_cmd_result.success and not overwrite:
-                    return grant_cmd_result
-            except JujuError as err:
-                if not overwrite:
-                    raise err
+            grant_cmd = GrantCommand()
+            grant_cmd_result: Result = await grant_cmd.run(
+                controller=controller, overwrite=overwrite, **kwargs
+            )
+            if not grant_cmd_result.success and not overwrite:
+                return grant_cmd_result
 
         return {
             "user": user.username,
--- a/juju_spell/commands/add_user.py
+++ b/juju_spell/commands/add_user.py
@@ -36,15 +36,23 @@ class AddUserCommand(BaseJujuCommand):
                 raise err
 
         enable_cmd = EnableUserCommand()
-        enable_cmd_result = await enable_cmd.run(controller=controller, **kwargs)
+        enable_cmd_result = await enable_cmd.run(
+            controller=controller, overwrite=overwrite, **kwargs
+        )
         if not enable_cmd_result.success and not overwrite:
             return enable_cmd_result
 
         if kwargs.get("acl"):
-            grant_cmd = GrantCommand()
-            grant_cmd_result: Result = await grant_cmd.run(controller=controller, **kwargs)
-            if not grant_cmd_result.success and not overwrite:
-                return grant_cmd_result
+            try:
+                grant_cmd = GrantCommand()
+                grant_cmd_result: Result = await grant_cmd.run(
+                    controller=controller, overwrite=overwrite, **kwargs
+                )
+                if not grant_cmd_result.success and not overwrite:
+                    return grant_cmd_result
+            except JujuError as err:
+                if not overwrite:
+                    raise err
 
         return {
             "user": user.username,
--- a/juju_spell/commands/add_user.py
+++ b/juju_spell/commands/add_user.py
@@ -2,6 +2,7 @@
 from typing import Any, Dict, Union
 
 from juju.controller import Controller
+from juju.errors import JujuError
 
 from juju_spell.commands.base import BaseJujuCommand, Result
 from juju_spell.commands.enable_user import EnableUserCommand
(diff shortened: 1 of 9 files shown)


chanchiwai-ray (commented):
LGTM

juju_spell/commands/add_user.py
```diff
@@ -23,6 +26,12 @@ async def execute(self, controller: Controller, *args: Any, **kwargs: Any) -> Di
             display_name=kwargs["display_name"],
         )
 
+        if kwargs.get("acl"):
+            grant_cmd = GrantCommand()
+            grant_result: Result = await grant_cmd.run(controller=controller, **kwargs)
+            if not grant_result.success:
+                return grant_result
```
rgildein, line 33: I think at this point we should raise an error and not return grand result.

juju_spell/commands/grant.py
```diff
@@ -14,21 +14,60 @@
...
         """Execute."""
-        result: bool = await controller.grant(username=kwargs["user"], acl=kwargs["acl"])
-        return result
+        acl = kwargs["acl"]
+        controller_acl: str
+        model_acl: str
+        if acl in CONTROLLER_ACL_CHOICES:
+            controller_acl = acl
+        else:
+            controller_acl = "login"
```
rgildein, line 48: We should not silently change the acl value here, since the validation is done in CLI it should never be anything else and if yes than it should raise an error.

rgildein: If we are trying to do some default values, than it should be like `if acl is None` or even better is `kwargs.get("acl", "login")`

juju_spell/commands/grant.py
```diff
@@ -14,21 +14,60 @@
...
+
+        if acl in MODEL_ACL_CHOICES:
+            model_acl = acl
+        elif acl == "superuser":
+            model_acl = "admin"
+        else:
+            model_acl = "read"
```
rgildein, line 55: similar situation here

juju_spell/commands/grant.py
```diff
@@ -14,21 +14,60 @@
...
+            models=models,
+            model_mappings=kwargs["controller_config"].model_mapping,
+        ):
+            grant_model_result: bool = await controller.grant_model(
+                username=kwargs["user"],
+                model_uuid=model.uuid,
+                acl=model_acl,
+            )
+            if not grant_model_result:
+                return grant_model_result
```
rgildein, line 72: There is no log message or any logic how differed between output from this process. I mean if i saw `False` I will not not why it's false. Can you some logs and maybe change it to return Result object with success = True/False and message which models was granted and which not? What do you think>?

tests/unit/commands/conftest.py
```diff
@@ -15,3 +16,9 @@ def test_juju_command():
     command = TestJujuCommand()
     command.execute.reset_mock()
     yield command
+
+
+async def _async_generator(values: List) -> AsyncGenerator:
```
rgildein, line 21: not-blocking: This is not a fixtures, so I'll suggest to move it to tests/unit/utils.py.

rgildein (changes requested):

juju_spell/commands/add_user.py
```diff
@@ -23,6 +26,12 @@ async def execute(self, controller: Controller, *args: Any, **kwargs: Any) -> Di
             display_name=kwargs["display_name"],
         )
 
+        if kwargs.get("acl"):
+            grant_cmd = GrantCommand()
```
esunar, line 30: Don't we also need enable_user and grand_model?

esunar (changes requested):

juju_spell/commands/grant.py
jneo8: > We should not silently change the acl value here, since the validation is done in CLI it should never be anything else and if yes than it should raise an error.

Here is the command, so we are not change the value.

Because we are grant user's acl to model and controller in the same time.

The acl value can be either model permission in [`read`, `write`, `admin`] or controller permission in [`superuser`, `login`]. The idea is user only need to type one of the acl then we will generate the correspond acl for user in both model and controller.

juju_spell/commands/grant.py
jneo8: We can add log msg here.

juju_spell/commands/add_user.py
jneo8: The grant model is been include in GrantCommand.

The enable_user is losing. I will add this.

juju_spell/commands/add_user.py
jneo8: Ack

juju_spell/commands/grant.py
rgildein: We should at-least moved this logic to separate function or make it simpler, because like this we have too many if statements.

juju_spell/commands/grant.py
```diff
@@ -14,21 +14,83 @@
...
+            controller_acl,
+            controller.uuid,
+            kwargs["user"],
+        )
+        result: bool = await controller.grant(username=kwargs["user"], acl=controller_acl)
+        if not result:
+            raise JujuSpellError(
+                f"Grant user {kwargs['user']} permission {controller_acl}"
+                f" on controller {controller.uuid} fail."
+            )
```
rgildein, line 72: I'm not sure if we should failed here or not. Maybe here yes and if grant to model failed not? What do you think @jneo8 and @esunar ? From user perspective.

jneo8: Updated.

After I checked the python-libjuju. It will raise error from the lib directly. So I remove the raise part here.

tests/unit/commands/conftest.py
jneo8: Ack

juju_spell/commands/add_user.py
rgildein: How this will help to user? You will return only `output = False` why?

juju_spell/commands/grant.py
```diff
@@ -14,21 +14,80 @@
...
+                model_acl,
+                controller.uuid,
+                kwargs["user"],
+            )
+            await controller.grant_model(
+                username=kwargs["user"],
+                model_uuid=model.uuid,
+                acl=model_acl,
+            )
+        return True
```
rgildein, line 93: It's not related to this PR, but maybe we could define that each commands should return Result object. In this case Something like Result(success=True, output="permission were granted for user %s in controller %s and following models %s"). What do you think @jneo8 and @esunar? Same to all commands, so we can easily aggregate results form multiple commands.

juju_spell/commands/grant.py
rgildein: Is that behavior which we want? If grant permission for single model failed (model is broken or I do not know) we will end up with user, which have permission only for some models

juju_spell/commands/grant.py
jneo8: Ack. unify log format is helpful.

juju_spell/commands/grant.py
jneo8: In the newest commit. We provide a argument `--overwrite`. It will ignore the exception and continue to next step, otherwise will get exception.

juju_spell/commands/grant.py
jneo8: Updated

juju_spell/commands/grant.py
jneo8: Updated

juju_spell/commands/add_user.py
jneo8: Not, the output will be `Result(success=False, error=${error-here})`.
Because the `run` function will return Result object, so it is simple logic to return the Result object from grant cmd.

juju_spell/commands/grant.py
```diff
@@ -14,21 +14,80 @@
...
-        return result
+        acl = kwargs["acl"]
+
+        controller_acl: str = self.get_controller_acl(acl)
+        model_acl: str = self.get_model_acl(acl)
+
+        logger.info(
+            "Start grant permission %s on controller %s for user %s",
+            controller_acl,
+            controller.uuid,
```
esunar, line 71: controller does not have uuid

juju_spell/commands/grant.py
```diff
@@ -14,21 +14,80 @@
...
+        await controller.grant(username=kwargs["user"], acl=controller_acl)
+
+        async for _, model in self.get_filtered_models(
+            controller=controller,
+            models=models,
+            model_mappings=kwargs["controller_config"].model_mapping,
+        ):
+            logger.info(
+                "Start grant model %s permission %s on controller %s for user %s",
+                model.uuid,
```
esunar, line 83: controller does not have uuid

esunar (changes requested):

juju_spell/commands/grant.py
esunar, line 71:
```suggestion
            controller.controller_uuid,
```

juju_spell/commands/grant.py
```diff
@@ -14,21 +14,80 @@
...
+                model_acl,
+                controller.uuid,
```
esunar, line 85:
```suggestion
                controller.controller_uuid,
```

esunar (approved):

juju_spell/commands/add_user.py
```diff
@@ -12,16 +15,44 @@
...
-            username=kwargs["user"],
-            password=password,
-            display_name=kwargs["display_name"],
+        try:
+            user = await controller.add_user(
+                username=kwargs["user"],
+                password=password,
+                display_name=kwargs["display_name"],
+            )
+        except JujuError as err:
```
rgildein, line 31: This is not good validation that user exist, because adding user could failed for other reason. You can check these [lines](https://github.com/juju/python-libjuju/blob/master/juju/controller.py#L649-L654). In fact it would be better to use `controller.get_user` instead of try and except here.

juju_spell/commands/add_user.py
```diff
@@ -12,16 +15,44 @@
...
-        user = await controller.add_user(
-            username=kwargs["user"],
-            password=password,
-            display_name=kwargs["display_name"],
+        try:
+            user = await controller.add_user(
+                username=kwargs["user"],
+                password=password,
+                display_name=kwargs["display_name"],
+            )
```
rgildein, line 30: we should add log message with information that user was added

juju_spell/commands/add_user.py
```diff
@@ -12,16 +15,44 @@
...
+        except JujuError as err:
+            if overwrite:
+                user = await controller.get_user(username=kwargs["user"])
+                await user.set_password(password)  # Reset user's password
```
rgildein, line 34: we should add log message that user password was changed

juju_spell/commands/add_user.py
```diff
@@ -12,16 +15,44 @@
...
+            try:
+                grant_cmd = GrantCommand()
+                grant_cmd_result: Result = await grant_cmd.run(
+                    controller=controller, overwrite=overwrite, **kwargs
+                )
+                if not grant_cmd_result.success and not overwrite:
+                    return grant_cmd_result
+            except JujuError as err:
+                if not overwrite:
+                    raise err
```
rgildein, line 55: Why this is needed? Can you provide some notes here, because we are not catching some specific error or can we make it more specific? The `controller.grant` is already doing some catching in these [lines](https://github.com/juju/python-libjuju/blob/master/juju/controller.py#L649-L654).

juju_spell/commands/enable_user.py
```diff
@@ -26,7 +27,17 @@
...
+        **kwargs: Any,
+    ) -> bool:
         """Execute."""
-        result: bool = await controller.enable_user(username=kwargs["user"])
-        return result
+        try:
+            await controller.enable_user(username=kwargs["user"])
+        except JujuError as err:
+            if not overwrite:
+                raise err
```
rgildein, line 42: Again this catching needs to be improved.

juju_spell/commands/enable_user.py
```diff
@@ -26,7 +27,17 @@
...
+        controller: Controller,
+        *args: Any,
+        overwrite: bool = False,
+        **kwargs: Any,
+    ) -> bool:
         """Execute."""
-        result: bool = await controller.enable_user(username=kwargs["user"])
-        return result
+        try:
+            await controller.enable_user(username=kwargs["user"])
```
rgildein, line 39: We should add some logs here.

juju_spell/commands/grant.py
```diff
@@ -14,21 +14,87 @@
...
+            "Start grant permission %s on controller %s for user %s",
+            controller_acl,
+            controller.controller_uuid,
+            kwargs["user"],
+        )
+        try:
+            await controller.grant(username=kwargs["user"], acl=controller_acl)
+        except JujuError as err:
+            if not overwrite:
+                raise err
```
rgildein, line 77: Again this exception is to generous, you do not know why it's failing.

rgildein (changes requested):

juju_spell/commands/add_user.py
rgildein: You change it, but before the grant command was not catching any error.

juju_spell/commands/grant.py
jneo8: grant model will fail if the permission is exists.

juju_spell/commands/add_user.py
jneo8: Ack. This one can be remove

juju_spell/commands/add_user.py
jneo8: Ack

juju_spell/commands/add_user.py
jneo8: Ack

juju_spell/commands/add_user.py
jneo8: Ack

juju_spell/commands/grant.py
rgildein: Yes it will failed if permission exists, but it can failed if your user could not grant permission or something went wrong with controller or ... . The `JujuError` could mean a lot of other things.

jneo8:
Reopen in #93

PR #90, opened 2023-03-06
feat(remove-user): remove user now disable user and revoke permission only

jneo8:


esunar (approved):

juju_spell/commands/enable_user.py
```diff
@@ -14,19 +14,32 @@
...
 """Command to enable users."""
-from typing import Any
+from typing import Any, Optional
 
 from juju.controller import Controller
 
 from juju_spell.commands.base import BaseJujuCommand
 
-__all__ = ["EnableUserCommand"]
+__all__ = ["EnableUserCommand", "DisableUserCommand"]
```
chanchiwai-ray, line 23: Maybe I missed the initial discussion about the code architeture. What is the reason for putting two commands in one single file ? Why can't we have separate file for disable user command ?

juju_spell/commands/enable_user.py
```diff
@@ -14,19 +14,32 @@
...
 
 class EnableUserCommand(BaseJujuCommand):
     """Enable user."""
 
-    async def execute(self, controller: Controller, *args: Any, **kwargs: Any) -> bool:
+    async def execute(self, controller: Controller, *args: Any, **kwargs: Any) -> None:
         """Execute."""
-        result: bool = await controller.enable_user(username=kwargs["user"])
-        return result
+        await controller.enable_user(username=kwargs["user"])
```
chanchiwai-ray, line 31: non-blocking: If enable user command now returns None, we may also want to update docstring in `cli.enable_user`.

juju_spell/commands/revoke.py
```diff
@@ -0,0 +1,56 @@
...
+        self,
+        controller: Controller,
+        *args: Any,
+        user: Optional[str] = None,
+        acl: Optional[str] = "login",
+        **kwargs: Any,
+    ) -> bool:
+        """Execute."""
+        await controller.revoke(username=user, acl=acl)
+        return True
```
chanchiwai-ray, line 39: Is it possible to return False? e.g. for a non-existing user.

juju_spell/commands/revoke.py
```diff
@@ -0,0 +1,56 @@
...
+        controller: Controller,
+        *args: Any,
+        user: Optional[str] = None,
+        model_uuid: Optional[str] = None,
+        acl: Optional[str] = "read",
+        **kwargs: Any,
+    ) -> bool:
+        """Execute."""
+        await controller.revoke_model(username=user, model_uuid=model_uuid, acl=acl)
+        return True
```
chanchiwai-ray, line 56: Is it possible to return False? e.g. for a non-existing user.

tests/unit/utils.py
```diff
@@ -0,0 +1,7 @@
+from typing import AsyncGenerator, List
+
+
+async def _async_generator(values: List) -> AsyncGenerator:
```
chanchiwai-ray, line 4: Naming suggestion: since it's used somewhere else (i.e. not private function), we should remove the prefix `_`.

chanchiwai-ray (changes requested):
Thanks for the patch, overall looks good to me, but I do left some non-blocking comments / suggestions.

juju_spell/commands/enable_user.py
rgildein: On the contrary, I think that we can move all commands for user to single file, so AddUser, RemoveUser, EnableUser, DisableUser, GrantUser. Otherwise we will have so many files with simple command like 5 lines EnableUserCommand in separate files.

juju_spell/commands/enable_user.py
rgildein: Why we are not returning anything here? Since `controller.enable_user` functions is returning something we should return that value.

juju_spell/commands/enable_user.py
```diff
@@ -14,19 +14,32 @@
...
+
+
+class DisableUserCommand(BaseJujuCommand):
+    """Disable user."""
+
+    async def execute(
+        self,
+        controller: Controller,
+        *args: Any,
```
rgildein, line 40: not-related: When I was adding mypy and pylint, I thought that we can change signature for BaseJujuCommand so function execute will not take any positional argument -> `*args: Any` will not be needed. What do you think?

juju_spell/commands/enable_user.py
```diff
@@ -14,19 +14,32 @@
...
 
-__all__ = ["EnableUserCommand"]
+__all__ = ["EnableUserCommand", "DisableUserCommand"]
 
 
 class EnableUserCommand(BaseJujuCommand):
     """Enable user."""
 
-    async def execute(self, controller: Controller, *args: Any, **kwargs: Any) -> bool:
+    async def execute(self, controller: Controller, *args: Any, **kwargs: Any) -> None:
```
rgildein, line 29: Can you also use `user: Optional[str] = None,` here.

juju_spell/commands/enable_user.py
```diff
@@ -14,19 +14,32 @@
...
+
+    async def execute(
+        self,
+        controller: Controller,
+        *args: Any,
+        user: Optional[str] = None,
+        **kwargs: Any,
+    ) -> None:
+        """Execute."""
+        await controller.disable_user(username=user)
```
rgildein, line 45: Again why not returning output from `controller.disable_user` function?

juju_spell/commands/enable_user.py
rgildein: But here it does not make sense to have `DisableUserCommand` in `enable_user.py`.

juju_spell/commands/revoke.py
rgildein: Again why we are not return result from `controller.revoke` function?

rgildein (changes requested):

juju_spell/commands/enable_user.py
jneo8: Maybe `user: str` is more exactly. It's required.

juju_spell/commands/enable_user.py
jneo8: After I checking the `python-libjuju`. The return data is nothing important.

jneo8:
Re-open in #93

Issue #89, opened 2023-03-03
Need to better architecture to accommodate caching in `juju-spell`.


chanchiwai-ray:
Background:

Sometimes controllers / models / applications information may not change very frequently, and connecting to controllers to fetch latest information can be very slow and unnecessary(e.g. you have many controllers, or controllers are far away from you physically) it will be ideal to have some "intelligent" caching mechanisms to cache those information, even without network connection.

Problems:

Running `juju-spell <command>` currently will always connect to controllers whether you're actually using it or not. If you just want a quick refresh of your knowledge about the clouds. This could take sometimes and it's not a good user experience. We might need to have a better architecture to accommodate the caching logic.

PR #92, opened 2023-03-08
Add caching feature to list models

chanchiwai-ray:
This MR add caching capability to `list-models` sub-command. It contains the following features:

  - It tries to load the cached result of `list-models` from `DEFAULT_CACHE_DIR`, and if not found, it will fetch result from the controller and then save the result to the cache directory.
  - If cached result exist, it use the cache result and prepend a warning message to instruct user to run `--refresh` if necessary.
  - **[reverted, will be in the future PR]** If cached result exist but the cache is expired (based on the cache policy), it will fetch result from the controller and then save the result to the cache directory.
- When run as `juju-spell list-models --refresh`
  -  It will fetch result from the controller and then save the result to the cache directory.

However, due to #89 , the caching feature currently does not improve the speed significantly.

### DIFF
--- a/juju_spell/cli/list_models.py
+++ b/juju_spell/cli/list_models.py
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Command entrypoint for ControllerInformationCommand."""
+"""Command entrypoint for ListModelsCommand."""
 import os
 import textwrap
 from datetime import datetime
--- a/juju_spell/cli/list_models.py
+++ b/juju_spell/cli/list_models.py
@@ -100,13 +100,15 @@ class ListModelsCMD(JujuReadCMD):
                 "output": {
                     "uuid": "03ascsb2-bba8-477b-854e-5715a7sb320a",
                     "name": "xxx-serverstack",
-                    "models": [
-                        "controller",
-                        "model-a",
-                        "model-b",
-                        "model-c"
-                    ],
-                    "refresh": true,
+                    "data": {
+                        "models": [
+                            "controller",
+                            "model-a",
+                            "model-b",
+                            "model-c"
+                        ],
+                        "refresh": true,
+                    },
                     "timestamp": 1078416000.0
                 },
                 "error": null
@@ -118,13 +120,14 @@ class ListModelsCMD(JujuReadCMD):
         controller_models_mapping = {}
 
         heading = ""
-        for outputs in retval:
-            controller_name = outputs["context"]["name"]
-            models = outputs["output"]["models"]
-            refresh = outputs["output"]["refresh"]
+        for content in retval:
+            output = content["output"]
+            controller_name = output["name"]
+            models = output["data"]["models"]
+            refresh = output["data"]["refresh"]
             controller_models_mapping[controller_name] = models
             if not refresh:
-                timestamp = float(outputs["output"]["timestamp"])
+                timestamp = float(output["timestamp"])
                 heading = WARNING_HEADING_TEMPLATE.format(
                     datetime.fromtimestamp(timestamp).strftime(DATETIME_FORMATTER)
                 )
--- a/juju_spell/cli/list_models.py
+++ b/juju_spell/cli/list_models.py
@@ -17,6 +17,7 @@
 """Command entrypoint for ControllerInformationCommand."""
 import os
 import textwrap
+from datetime import datetime
 from typing import Any
 
 import yaml
@@ -26,7 +27,10 @@ from craft_cli.dispatcher import _CustomArgumentParser
 from juju_spell.cli.base import JujuReadCMD
 from juju_spell.commands.list_models import ListModelsCommand
 
-WARNING_HEADING = "Use --refresh option with this command to see the latest information."
+WARNING_HEADING_TEMPLATE = (
+    "Use --refresh option with this command to see the latest information.\n\n(Last updated: {})"
+)
+DATETIME_FORMATTER = "%Y-%m-%d %H:%M:%S"
 
 
 class ListModelsCMD(JujuReadCMD):
@@ -52,7 +56,9 @@ class ListModelsCMD(JujuReadCMD):
         - model-c
 
         $ juju-spell list-models
-        "Use --refresh option with this command to see the latest information."
+        Use --refresh option with this command to see the latest information.
+
+        (Last updated: 1970-01-01 00:00:00)
 
         controller-a:
         - model-a
@@ -72,10 +78,7 @@ class ListModelsCMD(JujuReadCMD):
             "--refresh",
             default=False,
             action="store_true",
-            help=(
-                "This will force refresh all models from the controllers. (TODO: to be"
-                " implemented)"
-            ),
+            help="This will force refresh all models from the controllers.",
         )
 
     @staticmethod
@@ -94,13 +97,16 @@ class ListModelsCMD(JujuReadCMD):
                 },
                 "success": true,
                 "output": {
-                    "refresh": true,
+                    "uuid": "03ascsb2-bba8-477b-854e-5715a7sb320a",
+                    "name": "xxx-serverstack",
                     "models": [
                         "controller",
                         "model-a",
                         "model-b",
                         "model-c"
                     ],
+                    "refresh": true,
+                    "timestamp": 1172969203.1
                 },
                 "error": null
             }
@@ -116,7 +122,11 @@ class ListModelsCMD(JujuReadCMD):
             models = outputs["output"]["models"]
             refresh = outputs["output"]["refresh"]
             controller_models_mapping[controller_name] = models
-            heading = WARNING_HEADING if not refresh else ""
+            if not refresh:
+                timestamp = float(outputs["output"]["timestamp"])
+                heading = WARNING_HEADING_TEMPLATE.format(
+                    datetime.fromtimestamp(timestamp).strftime(DATETIME_FORMATTER)
+                )
 
         yaml_str = yaml.dump(
             controller_models_mapping,
--- a/juju_spell/cli/list_models.py
+++ b/juju_spell/cli/list_models.py
@@ -28,7 +28,8 @@ from juju_spell.cli.base import JujuReadCMD
 from juju_spell.commands.list_models import ListModelsCommand
 
 WARNING_HEADING_TEMPLATE = (
-    "Use --refresh option with this command to see the latest information.\n\n(Last updated: {})"
+    f"Use --refresh option with this command to see the latest information.{os.linesep*2}"
+    "(Last updated: {})"
 )
 DATETIME_FORMATTER = "%Y-%m-%d %H:%M:%S"
 
@@ -58,7 +59,7 @@ class ListModelsCMD(JujuReadCMD):
         $ juju-spell list-models
         Use --refresh option with this command to see the latest information.
 
-        (Last updated: 1970-01-01 00:00:00)
+        (Last updated: 2004-03-05 00:00:00)
 
         controller-a:
         - model-a
@@ -106,7 +107,7 @@ class ListModelsCMD(JujuReadCMD):
                         "model-c"
                     ],
                     "refresh": true,
-                    "timestamp": 1172969203.1
+                    "timestamp": 1078416000.0
                 },
                 "error": null
             }
--- a/juju_spell/commands/list_models.py
+++ b/juju_spell/commands/list_models.py
@@ -17,12 +17,12 @@ class ListModelsCommand(BaseJujuCommand):
         self, controller: Controller, *args: Any, **kwargs: Any
     ) -> Dict[str, Union[List, bool]]:
         """List models from the local cache or from the controllers."""
-        outputs: Union[JujuSpellError, Dict[str, Any]] = {}
+        outputs: Union[None, Dict[str, Any]] = {}
 
         if not kwargs["refresh"]:
             outputs = load_cache_data(kwargs["controller_config"].uuid, self.logger)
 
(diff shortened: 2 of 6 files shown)


juju_spell/commands/list_models.py
```diff
@@ -13,17 +17,51 @@ async def execute(
...
+
+def load_cache_data(uuid: str, logger: Logger) -> Union[JujuSpellError, Dict[str, Any]]:
+    """Gracefully load cache data from default cache directory."""
+    error_message_template = "%s list models failed to load cache: %s."
+    try:
+        cache_data = load_from_cache(uuid)
+        cache_data["refresh"] = False
+    except JujuSpellError as error:
+        logger.debug(error_message_template, uuid, str(error))
+        return JujuSpellError(error_message_template.format(uuid, str(error)))
```
rgildein, line 66: I think that this should raise an error or return None if cache does not exists.

juju_spell/commands/list_models.py
```diff
@@ -13,17 +17,51 @@ async def execute(
...
-        else:
-            models = ["TODO: read from cache; not implemented."]
+            outputs = save_cache_data(controller, models, kwargs["refresh"], self.logger)
 
-        outputs = {"refresh": kwargs["refresh"], "models": models}
         self.logger.debug("%s list models: %s", controller.controller_uuid, outputs["models"])
         return outputs
+
+
+def save_cache_data(
```
rgildein, line 39: I think that we should have this function predefined in units, so we can use it in multiple places and format will be same.
```python
cache_data = {
        "uuid": controller.controller_uuid,
        "name": controller.controller_name,
        "data": {
           "models": models,
           "refresh": refresh,
        },
        "timestamp": time(),
    }
```
What do you think?

rgildein (changes requested):
Two comments. I just want to confirmed that we change the design from using cached list of models (e.g. dealing with cache in `get_filtered_model_names`) to have cache for whole commands.

I think it's good idea, but we should move the mechanism to the BaseJujuCommand or to BaseJujuCachedCommand, so it will be easily re-usable. Not in this PR.

juju_spell/commands/list_models.py
chanchiwai-ray: ACK

juju_spell/cli/list_models.py
```diff
@@ -52,7 +56,9 @@ class ListModelsCMD(JujuReadCMD):
         - model-c
 
         $ juju-spell list-models
-        "Use --refresh option with this command to see the latest information."
+        Use --refresh option with this command to see the latest information.
+
+        (Last updated: 1970-01-01 00:00:00)
```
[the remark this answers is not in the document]
chanchiwai-ray: No reason, this date just come to my mind 😆 , change to the date Canonical founded.

juju_spell/cli/list_models.py
```diff
@@ -26,7 +27,10 @@
 from juju_spell.cli.base import JujuReadCMD
 from juju_spell.commands.list_models import ListModelsCommand
 
-WARNING_HEADING = "Use --refresh option with this command to see the latest information."
+WARNING_HEADING_TEMPLATE = (
+    "Use --refresh option with this command to see the latest information.\n\n(Last updated: {})"
```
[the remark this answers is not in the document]
chanchiwai-ray: ACK

juju_spell/commands/list_models.py
chanchiwai-ray: I think having a common format for `cached_data` is a good idea, but I currently do not have a complete picture of how are we going to use it (`cache_data`) in other places. I think we can keep it this way until there's more commands that can use `cache_data`.

chanchiwai-ray:
Thanks for the comments. Initially, I was thinking about only cache a list of models, but I realized that there are not many context about the models (e.g. from which controller, when did we create the cache, and how do we name the cache file, etc.). So, I decided to add more contexts, and in the end, it looks that I am caching for the whole commands (status, info, models) for a given controller.

I currently do not think we should put it `BaseJujuCommand` or deals with cache in `get_filtered_model_names`. This is because many commands could go wrong if we used an outdated `cache` (e.g. if we want to change user's permission of a model using the cache data, but the cache is out of date. It could cause command failed with weird errors, or if we want to list all units, but with a outdated cache...). I think we need more example of using the cache, and then refactor if needed.

juju_spell/commands/list_models.py
rgildein: Yes, that's true but since we do not have bigger picture and this is used only in one place it make sense to me to use one format in utils.
```python
def save_cache_data(
    controller: Controller, data: Dict[str, Any], logger: Logger
) -> Dict[str, Any]:
    cache_data = {
        "uuid": controller.controller_uuid,
        "name": controller.controller_name,
        "data": data,
        "timestamp": time(),
    }
```
Maybe if we will not be suitable for all scenarios, but that we can fix later and now it will be simpler + it will not force people to use a custom functions.

juju_spell/utils.py
```diff
@@ -96,3 +102,33 @@ def merge_list_of_dict_by_key(key: str, lists: List[List[Dict]]) -> List:
 def random_password(length: int = 30) -> str:
     """Generate random password."""
     return secrets.token_urlsafe(length)
+
+
+def save_to_cache(data: Dict, uuid: Union[str, Path]) -> None:
+    """Save data to the default cache directory named by the uuid."""
+    if not DEFAULT_CACHE_DIR.exists():
+        DEFAULT_CACHE_DIR.mkdir()
+    fname = DEFAULT_CACHE_DIR / uuid
```
rgildein, line 111: Now I check that we are using only controller uuid in file name, maybe we can use something like `<command_name>-<controller_uuid>`? It's not really needed in this PR.

rgildein:
> I currently do not think we should put it `BaseJujuCommand` or deals with cache in `get_filtered_model_names`. This is because many commands could go wrong if we used an outdated `cache` (e.g. if we want to change user's permission of a model using the cache data, but the cache is out of date. It could cause command failed with weird errors, or if we want to list all units, but with a outdated cache...). I think we need more example of using the cache, and then refactor if needed.

Yes I agree that it should not be in `BaseJujuCommand`, because we do not want to use cache for each command e.g. update-packages or add-user, but in future we can create `BaseReadJujuCommand`/`BaseCachecJujuCommand`, which will inherit from `BaseJujuCommand` and integrate cache mechanism. We will see, I'm just thinking at lout.

juju_spell/cli/list_models.py
```diff
@@ -116,7 +123,11 @@ def format_output(retval: Any) -> str:
             models = outputs["output"]["models"]
             refresh = outputs["output"]["refresh"]
             controller_models_mapping[controller_name] = models
-            heading = WARNING_HEADING if not refresh else ""
+            if not refresh:
+                timestamp = float(outputs["output"]["timestamp"])
+                heading = WARNING_HEADING_TEMPLATE.format(
+                    datetime.fromtimestamp(timestamp).strftime(DATETIME_FORMATTER)
```
esunar, line 129: I would prefer timestamp human readable

juju_spell/commands/list_models.py
```diff
@@ -13,17 +17,51 @@ async def execute(
...
+    controller: Controller, models: List[str], refresh: bool, logger: Logger
+) -> Dict[str, Any]:
+    """Gracefully save cache data to default cache directory."""
+    cache_data = {
+        "uuid": controller.controller_uuid,
+        "name": controller.controller_name,
```
esunar, line 45: I do not think we need name here. What will we do if name and uuid differs

juju_spell/utils.py
```diff
@@ -96,3 +102,33 @@ def merge_list_of_dict_by_key(key: str, lists: List[List[Dict]]) -> List:
...
+            file.write(data_string)
+    except PermissionError as error:
+        raise JujuSpellError(f"permission denied to write to file `{fname}`.") from error
+    except Exception as error:
+        raise JujuSpellError(f"{str(error)}.") from error
+
+
+def load_from_cache(uuid: Union[str, Path]) -> Dict[str, Any]:
+    """Load data from the default cache directory named by the uuid."""
+    fname = DEFAULT_CACHE_DIR / uuid
```
esunar, line 124: One of the biggest problems of using cache is cache eviction. I we are going to use caching application wise. It deserves own class with explicit policy. Something like:
```python
class JujuSpellCache:
    @abstractmethod
    def get(self, ) -> Any:
        pass

    @abstractmethod
    def put(self, cache) -> bool:
        pass

class CachingPolicy:
    policymap: Dict[str, int] = {} # "[command, timetolive in seconds]"
    ttl: int #timetolive in seconds

class FileCache(JujuSpellCache):
    policy: CachingPolicy

    def get(self, ) -> Any:
        pass

    def put(self, cache) -> bool:
        pass
```

juju_spell/cli/list_models.py
chanchiwai-ray: The data in the cache file and in python is timestamp, but when it's printed to the screen it's human readable like `2004-03-04 00:00:00`

juju_spell/commands/list_models.py
chanchiwai-ray: `uuid` is used to ID the cache file and provide more context about the cached data, and `name` is used when printing the message. E.g. 

```bash
<controller_name>
- model-a
- model -b
```

juju_spell/utils.py
chanchiwai-ray: I think policy like this is a good idea. It can be used to auto-refresh the cache if it's expired. Currently, the output message only gives you a warning like `use --refresh to get latest information ... (last updated <datetime>)`, and it's up to the user to refresh it. But with the policy, we can inspect the timestamp and decide if we want to automatically update it or not.

juju_spell/utils.py
chanchiwai-ray: Updated

juju_spell/utils.py
chanchiwai-ray: For now, I created simple `Cache` and `CachePolicy` to check if a cache is expired or not, but we can add more feature in the future if necessary.

juju_spell/commands/list_models.py
```diff
@@ -13,17 +17,55 @@ async def execute(
...
+            "refresh": refresh,
+        },
+        timestamp=time(),
+    )
+    fname = f"{command_name}_{controller.controller_uuid}"
+    error_message_template = "%s list models failed to save cache: %s."
+    try:
+        save_to_cache(cache, fname)
+    except JujuSpellError as error:
+        logger.debug(error_message_template, fname, str(error))
```
jneo8, line 58: **suggestion:** 

logger.warning/error?

Failing msg should be warning level.

juju_spell/utils.py
```diff
@@ -122,3 +176,33 @@ def load_yaml_file(path: Path) -> Any:
...
+
+
+def load_from_cache(name: Union[str, Path]) -> Cache:
+    """Load data from the default cache directory named by `name`."""
+    fname = DEFAULT_CACHE_DIR / name
+    try:
+        with open(fname, "r", encoding="UTF-8") as file:
+            data = yaml.safe_load(file)
+            return Cache(**data)
+    except FileNotFoundError as error:
```
jneo8, line 203: **suggestion:** 

I will consider to add log message here to tell the user why loading cache is fail. All the exception we know it may happen should have log.

jneo8 (changes requested):
Most LGTM. Need to add more log msg.

juju_spell/commands/list_models.py
chanchiwai-ray: Updated

juju_spell/utils.py
chanchiwai-ray: Added loggging when load cache is successful. But I think for the failing part, `craft_cli` will capute the error and we return that in the `Result.error`, so it might not be necessary.

juju_spell/cli/list_models.py
rgildein: Maybe @esunar mean that we should use human readable in cache file?

juju_spell/utils.py
```diff
@@ -15,19 +15,73 @@
...
 
 import yaml
 
 from juju_spell.exceptions import JujuSpellError
+from juju_spell.settings import DEFAULT_CACHE_DIR
 
 logger = logging.getLogger(__name__)
 
 
+class CachePolicy(dict):
```
rgildein, line 34: This should be in next PR.

juju_spell/utils.py
rgildein: We are raising from error so it's should be in log file.

juju_spell/utils.py
chanchiwai-ray: I agree, I also will put @esunar 's request in next PR about refactoring `save_to_cache` and `load_from_cache` function into a dedicated `FileCache` because it's quite a bit of changes.

juju_spell/utils.py
chanchiwai-ray: yes, it's in the log file if you check it with `--verbosity debug`

juju_spell/cli/list_models.py
esunar: I think so

chanchiwai-ray: It can, but I would opt to use a timestamp instead of human readable format. Because 1. the cache file is not intended to be read by human (although readble), 2. it's a lot easier to work with timestamp in python if you need to use the timestamp to do calculation e.g. checking if the cache is expired.

rgildein (approved):
LGTM

juju_spell/commands/list_models.py
rgildein: I still think that these function would be better only in utils, but since this is first command is not an issue.

PR #95, opened 2023-03-09
added changes requested during tests

esunar:
outputs converter from:
```
    "hostname": {
        "default": "",
        "description": "Override hostname, when empty uses default machine hostname",
        "source": "user",
        "type": "string",
        "value": "osman",
    },
    "xx": {
        "default": "",
        "description": "",
        "source": "user",
        "type": "string",
        "value": "xx",
    },
    "others": {
        "default": "",
        "description": "",
        "source": "user",
        "type": "string",
        "value": "xx",
    },
```

to:
`{"hostname": "osman", "xx": "xx", "others": "xx"}`

### DIFF
--- a/juju_spell/cli/config.py
+++ b/juju_spell/cli/config.py
@@ -40,7 +40,7 @@ class ConfigCMD(JujuWriteCMD):
         The config command gets or sets the config options on applications.
 
         Example:
-        $ juju-spell config --config-app nova-compute
+        $ juju-spell config nova-compute
         [
          {
           "context": {
@@ -50,13 +50,9 @@ class ConfigCMD(JujuWriteCMD):
           "output": {
             "my_model": {
               "my_application": {
-                "config_item": {
-                  "default": "",
-                  "description": "",
-                  "source": "",
-                  "type": "",
-                  "value": ""
-                },
+                "config_item": "config_value",
+                "config1_item": "config1_value",
+                "config2_item": "config2_value",
               }
             }
            }
@@ -65,8 +61,8 @@ class ConfigCMD(JujuWriteCMD):
          }
         ]
 
-        $ juju-spell config --config-app nova-compute --config-get config_item
-        $ juju-spell config --config-app nova-compute --config-set "a=1" "b=3"
+        $ juju-spell config nova-compute config_item
+        $ juju-spell config nova-compute --config-set "a=1" "b=3"
 
         With --file option multiple application configurations can be changed.
         $ juju-spell config --config-file config.yaml
--- a/juju_spell/cli/config.py
+++ b/juju_spell/cli/config.py
@@ -18,7 +18,6 @@
 import argparse
 import logging
 import textwrap
-from argparse import _MutuallyExclusiveGroup
 from pathlib import Path
 from typing import Any, List, Optional, Sequence, Union
 
@@ -88,9 +87,14 @@ class ConfigCMD(JujuWriteCMD):
         """Add arguments specific to the export-login command."""
         super().fill_parser(parser)
         parser.add_argument(
-            "--config-get",
-            required=False,
+            "config-app",
+            nargs="?",
+            help="The application to show/update configuration.",
+        )
+        parser.add_argument(
+            "config-get",
             type=str,
+            nargs="?",
             help="Single property to get.",
         )
         parser.add_argument(
@@ -100,24 +104,11 @@ class ConfigCMD(JujuWriteCMD):
             action=KeyValue,
             help="key=value pairs to update application config.",
         )
-
-        application_group = parser.add_mutually_exclusive_group(required=True)
-        self.add_mutually_exclusive_group_parameters(application_group)
-
-    def add_mutually_exclusive_group_parameters(
-        self, application_group: _MutuallyExclusiveGroup
-    ) -> None:
-        """Add mutually exclusive group."""
-        application_group.add_argument(
+        parser.add_argument(
             "--config-file",
             type=get_application_config,
             required=False,
             help="The path to yaml-formatted application config.",
-        )
-        application_group.add_argument(
-            "--config-app",
-            required=False,
-            help="The application to show/update configuration.",
         )
 
 
--- a/juju_spell/cli/config.py
+++ b/juju_spell/cli/config.py
@@ -84,7 +84,7 @@ class ConfigCMD(JujuWriteCMD):
     command = ConfigCommand
 
     def fill_parser(self, parser: _CustomArgumentParser) -> None:
-        """Add arguments specific to the export-login command."""
+        """Add arguments specific to the config command."""
         super().fill_parser(parser)
         parser.add_argument(
             "config-app",
--- a/juju_spell/commands/config.py
+++ b/juju_spell/commands/config.py
@@ -103,12 +103,9 @@ def values_only(
     """
     new_values: Dict[str, str] = {}
     for key, value in all_config_of_app.items():
-        new_values[key] = value.get("value")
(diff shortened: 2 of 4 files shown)


juju_spell/cli/config.py
```diff
@@ -88,9 +87,14 @@ def fill_parser(self, parser: _CustomArgumentParser) -> None:
         """Add arguments specific to the export-login command."""
```
chanchiwai-ray, line 87: This is not related to your patch, but maybe consider removing it since it's not about `export-login` command

juju_spell/commands/config.py
```diff
@@ -26,46 +28,90 @@ class ConfigCommand(BaseJujuCommand):
...
+    properties = [dns-ha]
+
+    to ->
+    {
+        "dns-ha": false,
+    }
+    """
+    new_values: Dict[str, str] = {}
+    for key, value in all_config_of_app.items():
+        new_values[key] = value.get("value")
```
chanchiwai-ray, line 106: non-blocking suggestion: maybe we can reduce a for-loop by combining the `if properties` logic in this loop. E.g.
```python
for key, value in all_config_of_app.items():
    if properties and key in properties:
        new_values[key] = value.get("value")
```

tests/unit/commands/test_cmd_config.py
```diff
@@ -8,18 +9,47 @@
...
-    def __init__(self, props: Dict[str, str]):
+    def __init__(self):
         """Initialize with dictionary."""
-        self.properties = props
+        self.properties = copy.deepcopy(FULL_CONFIG)
 
     async def set_config(self, props: Dict[str, str]):
-        self.properties.update(props)
+        for key, value in props.items():
+            print(key)
```
chanchiwai-ray, line 50: did you forget to remove it ?

chanchiwai-ray (changes requested):
LGTM overall, I have some small inline comments that you might want to apply.

juju_spell/cli/config.py
esunar: fixed

juju_spell/commands/config.py
esunar: I am leaving as is because I don't want to iterate if properties is empty.

tests/unit/commands/test_cmd_config.py
esunar: Removed

juju_spell/commands/config.py
chanchiwai-ray: I meant you can remove, if you use above logic.
```python
    if properties:
￼        for key in all_config_of_app.keys():
￼            if key not in properties:
￼                new_values.pop(key)
```
But i think it's not important which way you choose, since the complexity is just reduce from `O(2n) ~ O(n) to O(n)`

chanchiwai-ray (approved):

juju_spell/commands/config.py
esunar: If properties is null 
O(1)
else
O(n)

juju_spell/cli/config.py
```diff
@@ -85,12 +84,17 @@ class ConfigCMD(JujuWriteCMD):
     command = ConfigCommand
 
     def fill_parser(self, parser: _CustomArgumentParser) -> None:
-        """Add arguments specific to the export-login command."""
+        """Add arguments specific to the config command."""
         super().fill_parser(parser)
         parser.add_argument(
-            "--config-get",
-            required=False,
+            "config-app",
```
rgildein, line 90: I think you need also change example in overview.

juju_spell/commands/config.py
chanchiwai-ray:
```python3
    new_values: Dict[str, str] = {}
    for key, value in all_config_of_app.items(): # O(n)
        new_values[key] = value.get("value")

    if properties: # O(1) or O (n)
        for key in all_config_of_app.keys():
            if key not in properties:
                new_values.pop(key)
    return new_values

# total = O(n) + [O(1) or O(n)] = O(n)

---------------------------------------

    new_values: Dict[str, str] = {}
    for key, value in all_config_of_app.items(): # O(n)
        if properties and key in properties:
            new_values[key] = value.get("value")
    return new_values

# total = O(n)
```

rgildein:
```python
new_values: Dict[str, str] = {}
for key, value in all_config_of_app.items():
    if not properties or key in properties:
        new_values[key] = value.get("value")
```

rgildein (changes requested):
two small suggestion

juju_spell/commands/config.py
rgildein: I agree with @chanchiwai-ray here, but I think you need to use `if not properties or key in properties:`.

juju_spell/commands/config.py
```diff
@@ -26,46 +28,90 @@ class ConfigCommand(BaseJujuCommand):
...
     properties: Dict[str, str] = kwargs.get("config_set", None)
 
-    if application:
+    if updates:
+        config = await apply_file_config(model, updates)
+    elif application:
         juju_app: Application = model.applications.get(application)
         if not juju_app:
-            logger.warning("Unable to find application %s", application)
+            emit.message(f"model:{model.info.name} unable to find application:{application}")
```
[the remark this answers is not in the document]
esunar: message, verbose, debug, trace no other option

juju_spell/cli/config.py
esunar: done

juju_spell/commands/config.py
esunar: Sorry I understood a little late :(

esunar: fixed

juju_spell/commands/config.py
```diff
@@ -26,46 +28,90 @@ class ConfigCommand(BaseJujuCommand):
...
         else:
-            config[application] = all_config_of_app
+            config[application] = values_only(all_config_of_app, *keys)
 
-    elif updates:
-        config = await apply_file_config(model, updates)
     return config
 
 
+def values_only(
```
jneo8, line 76: **nitpick (non-blocking):** 

Maybe we can have more meaningful function name here.

jneo8 (approved):
Only one non-blocking comment

juju_spell/commands/config.py
esunar: Any suggestion?

juju_spell/commands/config.py
esunar: by the way `in` operator in also O(n) so the function is O(n^2)

juju_spell/commands/config.py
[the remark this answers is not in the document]
rgildein: There is also error, but no warning.... ok that's strange

juju_spell/cli/config.py
rgildein: Thanks

juju_spell/commands/config.py
jneo8: Something relate to serialization?

rgildein (approved):
LGTM

juju_spell/commands/config.py
esunar: nooo, because it has nothing to do with serialization

juju_spell/commands/config.py
chanchiwai-ray: you're right, `properies` is a tuple 😞 , I saw you use `in`, I thought it's `set` or `dict` like

PR #97, opened 2023-03-10
Fix emit.error

esunar:
Replaced emit.error with emit.message


rgildein (approved):
LGTM, Maybe it's really time to drop craft_cli, it would be very sad.

PR #98, opened 2023-03-10
remove args from abstract execute

rgildein:
There is really no need to use `*args` and execute function and mypy is forcing us to use it everywhere.


chanchiwai-ray (approved):
LGTM

esunar (approved):

PR #99, opened 2023-03-10
Add cache policy

chanchiwai-ray:
This PR extends the caching logic and defines a cache policy that can be used by any cache object. It's a conceptual implementation. Currently, we have a `FileCache` with a default cache policy `DefaultCachePolicy` that will check if the cache is expired or not, based on the default cache policy. So the cache logic is extended:

   - **[new]** If cached result exist but the cache is expired (based on the cache policy), it will fetch result from the controller and then save the result to the cache directory.

### DIFF
--- a/juju_spell/commands/list_models.py
+++ b/juju_spell/commands/list_models.py
@@ -1,13 +1,13 @@
 """List models from the local cache or from the controllers."""
 from logging import Logger
 from time import time
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Optional, Union
 
 from juju.controller import Controller
 
 from juju_spell.commands.base import BaseJujuCommand
 from juju_spell.exceptions import JujuSpellError
-from juju_spell.utils import Cache, load_from_cache, save_to_cache
+from juju_spell.utils import FileCache
 
 
 class ListModelsCommand(BaseJujuCommand):
@@ -17,13 +17,13 @@ class ListModelsCommand(BaseJujuCommand):
         self, controller: Controller, *args: Any, **kwargs: Any
     ) -> Dict[str, Union[List, bool]]:
         """List models from the local cache or from the controllers."""
-        cache: Union[None, Cache] = None
+        file_cache: Union[None, FileCache] = None
 
         if not kwargs["refresh"]:
             fname = f"{self.name}_{kwargs['controller_config'].uuid}"
-            cache = load_cache_data(fname, self.logger, kwargs["controller_config"].uuid)
+            file_cache = load_cache_data(fname, self.logger, kwargs["controller_config"].uuid)
 
-        if kwargs["refresh"] or cache is None:
+        if kwargs["refresh"] or file_cache is None or file_cache.expired:
             models = list(
                 await self.get_filtered_model_names(
                     controller=controller,
@@ -31,17 +31,21 @@ class ListModelsCommand(BaseJujuCommand):
                     model_mappings=kwargs["controller_config"].model_mapping,
                 )
             )
-            cache = save_cache_data(controller, models, kwargs["refresh"], self.logger, self.name)
+            file_cache = save_cache_data(
+                controller, models, kwargs["refresh"], self.logger, self.name
+            )
 
-        self.logger.debug("%s list models: %s", controller.controller_uuid, cache.data["models"])
-        return cache.context
+        self.logger.debug(
+            "%s list models: %s", controller.controller_uuid, file_cache.data["models"]
+        )
+        return file_cache.context
 
 
 def save_cache_data(
     controller: Controller, models: List[str], refresh: bool, logger: Logger, command_name: str
-) -> Cache:
+) -> FileCache:
     """Gracefully save cache data to default cache directory."""
-    cache = Cache(
+    file_cache = FileCache(
         uuid=controller.controller_uuid,
         name=controller.controller_name,
         data={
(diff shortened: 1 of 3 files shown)


juju_spell/utils.py
```diff
@@ -30,27 +34,131 @@
...
+        """Connect to the cache backend, need to implement by the subclass."""
+
+
+@dataclasses.dataclass()
+class FileCacheContext:
+    """A context for holding file cache data."""
+
+    uuid: str = ""
+    name: str = ""
+    data: Dict[str, Any] = dataclasses.field(default_factory=lambda: {})
```
rgildein, line 91: Why not using `dataclasses.field(default_factory=dict)`?

juju_spell/utils.py
```diff
@@ -30,27 +34,131 @@
 logger = logging.getLogger(__name__)
 
 
-@dataclasses.dataclass(frozen=True)
-class CacheContext:
-    """A context for holding cache data."""
+class CachePolicy(dict):
```
rgildein, line 37: not-blocking; We may consider to use `BaseCachePolicy` here.

rgildein: Also we may consider to use dataclass here instead of having dict. What do you think?

juju_spell/utils.py
```diff
@@ -30,27 +34,131 @@
...
+    def __init__(self, ttl: float, auto_refresh: bool = True):
+        """Initialize default cache policy."""
+        super().__init__(ttl=ttl, auto_refresh=auto_refresh)
 
-class Cache(CacheContext):
+
+class Cache(metaclass=ABCMeta):
     """A cache for command's output with cache policy."""
 
+    policy: DefaultCachePolicy = DefaultCachePolicy(ttl=3600, auto_refresh=True)
```
rgildein, line 60: define constants for ttl and auto_refresh

juju_spell/utils.py
```diff
@@ -30,27 +34,131 @@
...
+    @abstractmethod
+    def get(self, key: str) -> Any:
+        """Get data from the cache, need to implement by the subclass."""
+
+    @abstractmethod
+    def put(self, key: str, value: Any) -> None:
+        """Update the cache's data, need to implement by the subclass."""
+
+    @abstractmethod
+    def commit(self) -> Any:
```
rgildein, line 76: What will commit return here?

juju_spell/utils.py
```diff
@@ -30,27 +34,131 @@
...
+        """Commit the changes to the cache, need to implement by the subclass."""
+
+    @classmethod
+    @abstractmethod
+    def connect(cls, backend: Any) -> Any:
+        """Connect to the cache backend, need to implement by the subclass."""
+
+
+@dataclasses.dataclass()
+class FileCacheContext:
```
rgildein, line 86: Is there any particular reason why this could not be part of FileCache object?

juju_spell/utils.py
```diff
@@ -30,27 +34,131 @@
...
-        """Return the cache context as a dictionary."""
+        """Get the file cache context as a dictionary."""
         return dataclasses.asdict(self)
 
-    def add_policy(self) -> None:
-        """Add more policy to the default policy."""
-        raise NotImplementedError("Add policy is not supported yet.")
+    def put(self, key: str, value: Any) -> None:
+        """Change file cache data."""
+        setattr(self, key, value)
```
rgildein, line 113: Like this user can save anything to the cache and I think we should not enable it, because format of file should be same for any cache. Save any output to data is ok, but saving output to `output` or `result`, ... it's should not work.

juju_spell/utils.py
```diff
@@ -30,27 +34,131 @@
...
+    def commit(self, name: Optional[Union[str, Path]] = "") -> None:
+        """Commit file cache to the default cache directory named by `name`."""
+        if not self.cache_directory.exists():
+            self.cache_directory.mkdir()
+        if self.cache_name == "" and name == "":
+            raise JujuSpellError(
+                "failed to write commit file cache: missing 'name' for file cache."
+            )
+        if name:
+            self.cache_name = name
```
rgildein, line 133: not-blocking
```suggestion
        self.cache_name = name or self.cache_name
        if self.cache_name == "":
            raise JujuSpellError(
                "failed to write commit file cache: missing 'name' for file cache."
            )
```

juju_spell/utils.py
```diff
@@ -30,27 +34,131 @@
...
+                data = yaml.safe_dump(self.context)
+                logger.debug("save cache file to %s", str(fname))
+                file.write(data)
+        except PermissionError as error:
+            raise JujuSpellError(f"permission denied to write to file `{fname}`.") from error
+        except Exception as error:
+            raise JujuSpellError(f"{str(error)}.") from error
+
+    @classmethod
+    def connect(cls, backend: Union[str, Path]) -> FileCache:
```
rgildein, line 146: Would not be better to use `__init___`? So I can do `file_cache = FileCache(fname)` instead of `file_cache = FileCache.connect(fname)`.

juju_spell/commands/list_models.py
```diff
@@ -17,31 +17,35 @@ async def execute(
...
 
-        self.logger.debug("%s list models: %s", controller.controller_uuid, cache.data["models"])
-        return cache.context
+        self.logger.debug(
+            "%s list models: %s", controller.controller_uuid, file_cache.data["models"]
+        )
+        return file_cache.context
 
 
 def save_cache_data(
```
rgildein, line 44: Since we are passing `self.logger` and `self.name` why not put these functions (save_cache_data and save_cache_data) to ListModelsCommand. Would not be better?

juju_spell/utils.py
chanchiwai-ray: ACK

juju_spell/utils.py
chanchiwai-ray: ACK: `BaseCachePolicy`. For `dataclass` over `dict`, my original consideration here is that we might want to combine default policy (hard-coded) with a user provided (availiable / supported) policy file, so `dict` can reflect the dynamics of this. However, after a second thought, I think `dataclass` is also okay, if we pre-define all available policy (filled with `None` for example) can also achieve the same goal, and it might makes the code look nicer.

juju_spell/utils.py
chanchiwai-ray: ACK

juju_spell/utils.py
chanchiwai-ray: it's a typo, should be `None`

juju_spell/utils.py
chanchiwai-ray: it's a syntatic sugar, we can easily get the `data` through `dataclasses.asdict(self)` and it will not contain any properies defined in `FileCache`

juju_spell/utils.py
chanchiwai-ray: Actually, it won't be able to since whatever being `committed` is only in `self.context` and that `context` is comming from the `FileCacheContext` which is the same for any cache. It's also related to your above question.

juju_spell/utils.py
chanchiwai-ray: 1. This is because of `FileCacheContext` mixin class. Using `__init__` will need to supply the data itself when `connect(fname, **data)` which we don't have initially before doing any operation (this is true even we don't use mixin class and put everything in `FileCache` class). 
2. For consistency, as @esunar mentioned, we might want to implement different kind of cache backend e.g. a DB, using `connect` as a abstract classemethod can force / remind user to implment that, and it shares the same API (though we can improve the share API if it's not good).

juju_spell/commands/list_models.py
chanchiwai-ray: We could..., I just don't think the code will "look" better, because it will be a very **big** `if` in the `ListModelsCommand`. Or did I miss understand what you mean? Did you mean as a `method` in `ListModelsCommand`?

chanchiwai-ray:
I will close this MR because it will need to rework

PR #100, opened 2023-03-14
feat(juju_spell/cmd.py): Print traceback_format_exc to major error handler when err is exception or JujuSpellError

jneo8:
Only print traceback_format_exc if the EmitterMode is either BRIEF or QUIET

The craft_cli error will print traceback mesage if the verbosity is either `DEBUG` or `TRACE`.

In this PR I think we should handle JujuSpellError just like the normal exception. So it will output the traceback msg automatically in std and file.
But sometime we need to debug with the traceback msg. So I add simple print when the verbosity is <= `VERBOSE`.  This will not been record in file output but provide us more information to debug.

Example:

Adding `raise Exception("testing")`

```python
def exec_cmd() -> int:
    """Execute craft cli."""
    dispatcher = get_dispatcher()
    return_code = 0

    try:
        _run_dispatcher(dispatcher)
        raise Exception("testing")
        emit.ended_ok()
    except ArgumentParsingError as err:

```

```sh
$ python -m juju_spell ping 
[
 {
  "context": {
   "uuid": "9d3fb816-fb10-461a-8230-0845cb8afa11",
   "name": "remote1-local-lxc",
   "customer": "remote1"
  },
  "success": true,
  "output": "accessible",
  "error": null
 }
]                             
Traceback (most recent call last):
  File "/home/jneo8/CanonicalProjects/juju-spell/juju_spell/cmd.py", line 155, in exec_cmd
    raise Exception("testing")
Exception: testing


Application internal error: Exception('testing')                                                                                                                                                                                   
Full execution log: '/home/jneo8/.local/state/juju-spell/log/juju-spell-20230314-160526.252198.log' 
```


juju_spell/cmd.py
```diff
@@ -152,29 +153,27 @@ def exec_cmd() -> int:
...
         craft_error = CraftError("Interrupted.")
-        craft_error.__cause__ = exc
+        craft_error.__cause__ = err
         emit.error(craft_error)
         return_code = 130
-    except Exception as exc:  # pylint: disable=W0718
-        craft_error = CraftError(f"Application internal error: {exc!r}")
-        craft_error.__cause__ = exc
+    except (JujuSpellError, Exception) as err:  # pylint: disable=W0718
+        if emit._mode in [EmitterMode.BRIEF, EmitterMode.QUIET]:  # pylint: disable=W0212
```
rgildein, line 173: Is there reason why we are not using `emit.get_mode()`?

rgildein:
Can you provide examples how it look like in real world? Raise exception in single command to see it in CLI output.
I'm not sure if `JujuSpellError` should be handled as normal exception, maybe return_code could be 1 instead of 0, but for example having `Application internal error: configuration file validation failed` instead of `configuration file validation failed` it's not nice, I think.

esunar (commented):
Can you provide sample error output?

juju_spell/cmd.py
jneo8: The craft-cli will not output traceback is the verbosity is BRIEF and QUIET. I think the question here is like *Do we want to provide traceback msg even `--verbose` is not provided*

rgildein: No question here is why we are using `emit._mode ` instead of `emit.get_mode()`.

jneo8:
Update some details.

Fix the Emit level, right now will output traceback in all level.

esunar (approved):

juju_spell/cmd.py
jneo8: Updated to get_mode()

esunar (approved):

test.json
```diff
@@ -0,0 +1,3 @@
+{
```
rgildein, line 1: What is this file for?

jneo8: Removed

PR #102, opened 2023-03-15
Create generic caching mechanism

chanchiwai-ray:
This PR extends and generailized the existing caching mechanism and defines a cache policy that can be used by any cache object.

   - **[new]** If cached result exist but the cache is expired (based on the cache policy), it will fetch result from the controller and then save the result to the cache directory.
- Define abstract generic interface `get` and `put`
- Implement concrete `FileCache` with `get`, `put`, and `delete`

Example cache layout:

```bash
/home/raychan/.local/share/juju-spell/caches/
├── ListModelsCommand_030ccab2-b4a8-477b-854e-5715a74b320a
└── ListModelsCommand_bf74ae2d-5934-4adf-86c9-3393882682c8
```

Example cache file content:

```bash
# ~/.local/share/juju-spell/caches/ListModelsCommand_030ccab2-b4a8-477b-854e-5715a74b320a
data:
  models:
  - controller
  - cos-lite
  - jba
  - raychan96
timestamp: 1678870440.5674603
```

Reopen #99

### DIFF
--- a/juju_spell/cache.py
+++ b/juju_spell/cache.py
@@ -33,7 +33,7 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_TTL_RULE = 3600  # in seconds
 DEFAULT_CACHE_BACKEND = "FileCache"
-AVALIABLE_BACKENDS = {"FileCache"}
+AVALIABLE_BACKENDS = {DEFAULT_CACHE_BACKEND}
 
 
 def use_cache(backend: str = DEFAULT_CACHE_BACKEND) -> "Cache":
--- a/juju_spell/cache.py
+++ b/juju_spell/cache.py
@@ -1,0 +1,141 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021-2022 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Data structures to offer caching cache to juju-spell."""
+from __future__ import annotations
+
+import dataclasses
+import logging
+from abc import ABCMeta, abstractmethod
+from pathlib import Path
+from time import time
+from typing import Any, Dict, Optional
+
+import yaml
+
+from juju_spell.exceptions import JujuSpellError
+from juju_spell.settings import DEFAULT_CACHE_DIR
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TTL_RULE = 3600  # in seconds
+DEFAULT_CACHE_BACKEND = "FileCache"
+AVALIABLE_BACKENDS = {"FileCache"}
+
+
+def use_cache(backend: str = DEFAULT_CACHE_BACKEND) -> "Cache":
+    """Get the cache class based on the caching backend."""
+    if backend not in AVALIABLE_BACKENDS:
+        raise JujuSpellError(f"Supported cache backends are {AVALIABLE_BACKENDS}")
+    return FileCache()
+
+
+@dataclasses.dataclass
+class CachePolicy:
+    """A class for cache policy."""
+
+    ttl: int = DEFAULT_TTL_RULE
+
+
+class Cache(metaclass=ABCMeta):
+    """A cache for command's output with cache policy."""
+
+    policy: CachePolicy = CachePolicy()
+    cache_directory: Path = DEFAULT_CACHE_DIR
+
+    def __init__(self) -> None:
+        """Initialize the cache."""
+        if not self.cache_directory.exists():
+            self.cache_directory.mkdir()
+
+    @abstractmethod
+    def get(self, key: str) -> Any:
+        """Get data from the cache, need to implement by the subclass."""
+
+    @abstractmethod
+    def put(self, key: str, value: Any) -> None:
+        """Update the cache's data, need to implement by the subclass."""
+
+
+@dataclasses.dataclass
+class FileCacheContext:
+    """A context for holding output from each command."""
+
+    timestamp: float = time()
+    data: Dict[str, Any] = dataclasses.field(default_factory=dict)
+
+
+class FileCache(Cache):
+    """Create a cache file to store the data."""
+
+    def get(self, key: str) -> Optional[Dict[str, Any]]:
+        """Get the data from the file cache."""
+        context = self._connect(key)
+        if not context:
+            return None
+        return dataclasses.asdict(context)
+
+    def put(self, key: str, value: Dict[str, Any]) -> None:
+        """Add or update the data in the file cache."""
+        self._commit(key, FileCacheContext(timestamp=time(), data=value))
+
+    def delete(self, key: str) -> None:
+        """Delete the data from the file cache."""
+        self._remove(key)
+
+    def check_expired(self, key: str) -> bool:
+        """Check if the cache is expired or not."""
+        context = self._connect(key)
+        if not context:
+            return True
+        return context.timestamp + self.policy.ttl < time()
+
+    def _commit(self, key: str, context: FileCacheContext) -> None:
+        """Commit the changes to the disk."""
+        cache_name = self.cache_directory / key
+        try:
+            with open(cache_name, "w", encoding="UTF-8") as file:
+                data = yaml.safe_dump(dataclasses.asdict(context))
+                file.write(data)
+        except PermissionError as error:
+            raise JujuSpellError(f"permission denied to write to file `{cache_name}`.") from error
+        except Exception as error:
+            raise JujuSpellError(f"{str(error)}.") from error
+
+    def _connect(self, key: str) -> Optional[FileCacheContext]:
+        """Connect to file cache or create a new file cache."""
+        cache_name = self.cache_directory / key
+        try:
+            if not cache_name.exists():
+                return None
+            with open(cache_name, "r", encoding="UTF-8") as file:
+                return FileCacheContext(**yaml.safe_load(file))
+        except PermissionError as error:
+            raise JujuSpellError(f"permission denied to read from file `{cache_name}`.") from error
+        except Exception as error:
+            raise JujuSpellError(f"{str(error)}.") from error
+
+    def _remove(self, key: str) -> None:
+        """Remove the data from the file cache."""
+        cache_name = self.cache_directory / key
+        try:
+            cache_name.unlink()
+        except FileNotFoundError as error:
+            raise JujuSpellError(f"`{cache_name}` does not exists.") from error
+        except PermissionError as error:
+            raise JujuSpellError(f"permission denied to read from file `{cache_name}`.") from error
+        except Exception as error:
+            raise JujuSpellError(f"{str(error)}.") from error
--- a/juju_spell/cache.py
+++ b/juju_spell/cache.py
@@ -63,11 +63,15 @@ class Cache(metaclass=ABCMeta):
 
     @abstractmethod
     def get(self, key: str) -> Any:
-        """Get data from the cache, need to implement by the subclass."""
+        """Get the data from the cache."""
 
     @abstractmethod
     def put(self, key: str, value: Any) -> None:
-        """Update the cache's data, need to implement by the subclass."""
+        """Add or update the data in the cache."""
+
+    @abstractmethod
+    def delete(self, key: str) -> None:
+        """Delete the data from the cache."""
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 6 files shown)


juju_spell/cache.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @abstractmethod
+    def get(self, key: str) -> Any:
+        """Get data from the cache, need to implement by the subclass."""
+
+    @abstractmethod
+    def put(self, key: str, value: Any) -> None:
+        """Update the cache's data, need to implement by the subclass."""
+
+    def update(self, **kwargs: Any) -> None:
```
esunar, line 65: We should not have such method

juju_spell/cache.py
```diff
@@ -0,0 +1,147 @@
...
+
+DEFAULT_TTL_RULE = 3600  # in seconds
+AVALIABLE_BACKENDS = {"FileCache"}
+
+
+def get_cache(backend: str = "FileCache") -> Any:
+    """Get the cache class based on the caching backend."""
+    if backend not in AVALIABLE_BACKENDS:
+        raise JujuSpellError(f"Supported cache backends are {AVALIABLE_BACKENDS}")
+    return FileCache
```
esunar, line 42: factory should return instance not type

juju_spell/commands/list_models.py
```diff
@@ -17,59 +15,28 @@ async def execute(
         self, controller: Controller, *args: Any, **kwargs: Any
     ) -> Dict[str, Union[List, bool]]:
         """List models from the local cache or from the controllers."""
-        cache: Union[None, Cache] = None
+        cache = get_cache()(controller.controller_uuid, controller.controller_name)
```
esunar, line 18: cache should be created when the object is created

esunar (changes requested):
Overall structure is settling, however I cannot see an expire or delete invalidate function in cache for removing objects from cache manually.

juju_spell/cache.py
```diff
@@ -0,0 +1,147 @@
...
+from juju_spell.exceptions import JujuSpellError
+from juju_spell.settings import DEFAULT_CACHE_DIR
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TTL_RULE = 3600  # in seconds
+AVALIABLE_BACKENDS = {"FileCache"}
+
+
+def get_cache(backend: str = "FileCache") -> Any:
```
rgildein, line 38: I think that it would be better to use constant for `"FileCache"`, so it can be used in multiple places.

juju_spell/commands/list_models.py
rgildein, line 18: I think it would be better to not use `FileCache` as default option or at least specify it here with `get_cache(FILE_CACHE)(...)`. What do you think?

juju_spell/cache.py
rgildein, line 38:
```suggestion
def get_cache(backend: str = "FileCache") -> "Cache":
```

juju_spell/cache.py
chanchiwai-ray: ACK

juju_spell/cache.py
chanchiwai-ray: ACK

juju_spell/commands/list_models.py
chanchiwai-ray: I agree. The expected behavour should be like `_stored` in `ops` framework where we create the cache at the very beginning of the app, and pass it down to `async def execute(self, controller, cache):`. But now we don't really have this "beginning", so I put it here. We should revisit this after we refactor the code, or do you have any suggestion?

rgildein (changes requested):
Couple of small comments, but overall LGTM. I tested it and it's works as intended, however I'm little afraid that we are writing everything to same file and Cache is not designer to have just one connection. What could happens is that multiple commands will write to same file at the same time. Not sure if it could be possible :thinking:. What do you think about using
```bash
.local/share/juju-spell/caches
├── list-models
│   ├── e9fe93a1-b705-4067-8f30-6eec183eeb4f
│   └── e9fe93a8-b705-4067-8f30-6eec183eeb4f
```

juju_spell/commands/list_models.py
chanchiwai-ray: I was thinking, after we finish the refactor, this parameter is given by config or cli args, and if they don't provide it, config or cli args should gives the default value. Anyways, this line should be revisited after refactoring since we don't want to create cache instance for each command.

chanchiwai-ray:
> Couple of small comments, but overall LGTM. I tested it and it's works as intended, however I'm little afraid that we are writing everything to same file and Cache is not designer to have just one connection. What could happens is that multiple commands will write to same file at the same time. Not sure if it could be possible thinking. What do you think about using
> 
> ```shell
> .local/share/juju-spell/caches
> ├── list-models
> │   ├── e9fe93a1-b705-4067-8f30-6eec183eeb4f
> │   └── e9fe93a8-b705-4067-8f30-6eec183eeb4f
> ```

I guess it won't... since the cache file is per controller, and within the controller, the command is sequential, and each `put` or `get` transaction is atomic. So, it shouldn't be read / write by multiple commands. But if it is problem, maybe we need to add locks to the cache.

rgildein:
> I guess it won't... since the cache file is per controller, and within the controller, the command is sequential, and each `put` or `get` transaction is atomic. So, it shouldn't be read / write by multiple commands at the same time. But if it is problem, maybe we need to add locks to the cache.

Yes, but same controller can be defined twice in one config or one command is running, like upgrade-packages, and in mean time I want to get list of models or maybe status of models. So There are scenarios where we could write to same file at once.

chanchiwai-ray:
> > I guess it won't... since the cache file is per controller, and within the controller, the command is sequential, and each `put` or `get` transaction is atomic. So, it shouldn't be read / write by multiple commands at the same time. But if it is problem, maybe we need to add locks to the cache.
> 
> Yes, but same controller can be defined twice in one config or one command is running, like upgrade-packages, and in mean time I want to get list of models or maybe status of models. So There are scenarios where we could write to same file at once.

In this case, we probably want to use `threading.Lock()` instead of spliting multiple caching files. Because otherwise we will need to create the cache object ourself inside the command, which may not be a good developing experience.

juju_spell/cache.py
chanchiwai-ray: ACK

chanchiwai-ray: ACK

rgildein:
In `sqlite` it would not be much problem, because you will have separate database for each command and here my suggestion is only to not store everything in one file, that's all.

juju_spell/cache.py
chanchiwai-ray: If I define this, and add a property in `FileCache`, mypy does not like it 
```
lint run-test: commands[2] | mypy .
juju_spell/commands/list_models.py:29: error: "Cache" has no attribute "expired"  [attr-defined]
```

juju_spell/commands/list_models.py
chanchiwai-ray: ACK

chanchiwai-ray:
> In `sqlite` it would not be much problem, because you will have separate database for each command and here my suggestion is only to not store everything in one file, that's all.

Okay, that make sense. I've updated it to create cache directory per command and per controller

juju_spell/cache.py
```diff
@@ -0,0 +1,141 @@
...
+
+@dataclasses.dataclass
+class CachePolicy:
+    """A class for cache policy."""
+
+    ttl: int = DEFAULT_TTL_RULE
+
+
+class Cache(metaclass=ABCMeta):
+    """A cache for command's output with cache policy."""
```
esunar, line 54: Still we do not have a way to remove item from cache.

esunar (changes requested):
Only missing remove. Can you look [this](https://pastebin.canonical.com/p/kCbvn3ZMY9/) as reference. This is 2 interface definitions from well known java spring framework.

juju_spell/cache.py
chanchiwai-ray: I implemented it in the `FileCache`, but did not specify in `Cache`

juju_spell/commands/list_models.py
```diff
@@ -17,59 +14,27 @@ async def execute(
         self, controller: Controller, *args: Any, **kwargs: Any
     ) -> Dict[str, Union[List, bool]]:
         """List models from the local cache or from the controllers."""
-        cache: Union[None, Cache] = None
+        cache: Any = use_cache(DEFAULT_CACHE_BACKEND)
```
rgildein, line 17:
```suggestion
        cache: Cache = use_cache(DEFAULT_CACHE_BACKEND)
```

chanchiwai-ray:
```
lint run-test: commands[2] | mypy .
juju_spell/commands/list_models.py:28: error: "Cache" has no attribute "check_expired"  [attr-defined]
Found 1 error in 1 file (checked 42 source files)
```
Do you know how to fix it?

rgildein (commented):
Is this right approach? If cache does not exist we are pretending that cache exists.
```bash
(.venv) x1:➜  juju-spell git:(add-generic-cache-mechanism) rm -rf ~/.local/share/juju-spell/caches
(.venv) x1:➜  juju-spell git:(add-generic-cache-mechanism) ll ~/.local/share/juju-spell/            
total 8.0K
-rwxr-xr-x 1 rgildein rgildein 5.7K Mar 10 07:33 config.yaml                                                                         [0.00s]
(.venv) x1:➜  juju-spell git:(add-generic-cache-mechanism) juju-spell list-models --filter owner=Frodo
Use --refresh option with this command to see the latest information.

(Last updated: 2023-03-16 08:17:48)

u1:
- controller
- cos-lite
- default
- exporters
- lma
u2:
- controller
- cos-lite
- default
- exporters
- lma                                                                                                                                [3.94s]
```

not-blocking: Also I think it would be better to have `list-models` sub-directory instead of prefix in file name. 
`./ListModelsCommand_e9fe93a8-b705-4067-8f30-6eec183eeb4f` -> `./ListModelsCommand/e9fe93a8-b705-4067-8f30-6eec183eeb4f`.

juju_spell/commands/list_models.py
rgildein: `check_expired` should be in `Cache`? Or we only expect `FileCache` here? If yes, it should be type `FileCache` and not `Any`.

chanchiwai-ray:
> Is this right approach? If cache does not exist we are pretending that cache exists.
> 
> ```shell
> (.venv) x1:➜  juju-spell git:(add-generic-cache-mechanism) rm -rf ~/.local/share/juju-spell/caches
> (.venv) x1:➜  juju-spell git:(add-generic-cache-mechanism) ll ~/.local/share/juju-spell/            
> total 8.0K
> -rwxr-xr-x 1 rgildein rgildein 5.7K Mar 10 07:33 config.yaml                                                                         [0.00s]
> (.venv) x1:➜  juju-spell git:(add-generic-cache-mechanism) juju-spell list-models --filter owner=Frodo
> Use --refresh option with this command to see the latest information.
> 
> (Last updated: 2023-03-16 08:17:48)
> 
> u1:
> - controller
> - cos-lite
> - default
> - exporters
> - lma
> u2:
> - controller
> - cos-lite
> - default
> - exporters
> - lma                                                                                                                                [3.94s]
> ```

Fixed, it's just the formatting issue, the content is fetched from the controller if it does not has the cache.


> not-blocking: Also I think it would be better to have `list-models` sub-directory instead of prefix in file name. `./ListModelsCommand_e9fe93a8-b705-4067-8f30-6eec183eeb4f` -> `./ListModelsCommand/e9fe93a8-b705-4067-8f30-6eec183eeb4f`.

The file name is simply the `key` in `cache.put(key, data)`. Currently, I don't have a good idea how to originize the cache directory (I think it's not important because normally it's not expected to be seen by end users) without breaking the `Cache` interface. i.e.

```python
# Expected cache interface
------------------------------------
# somewhere not in the `command.execute` and maybe globally defined `from ... import cache`
cache = use_cache(some_backend)
------------------------------------
# in the `command.execute`
def execute(self, controller):
    ...
    key = f"{self.name}_{controller.controller_uuid}"
    cache.put(key, data)
    data = cache.get(key)
    cache.delete(key) # remove the cache file
    ...

# if we organize the cache directory, the cache should somehow know the command name to begin with
------------------------------------
# in the `command.execute`
def execute(self, controller):
    key = f"{controller.controller_uuid}"
    cache = use_cache(some_backend, self.name)
    cache.put(key, data)
    data = cache.get(key)
    cache.delete(key) # remove the cache file

```

juju_spell/commands/list_models.py
chanchiwai-ray: Added to `Cache` as a optional method

juju_spell/cache.py
```diff
@@ -0,0 +1,151 @@
...
+import yaml
+
+from juju_spell.exceptions import JujuSpellError
+from juju_spell.settings import DEFAULT_CACHE_DIR
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TTL_RULE = 3600  # in seconds
+DEFAULT_CACHE_BACKEND = "FileCache"
+AVALIABLE_BACKENDS = {"FileCache"}
```
rgildein, line 36: not-blocking: 
```suggestion
AVALIABLE_BACKENDS = {DEFAULT_CACHE_BACKEND}
```

esunar (approved):

rgildein (approved):
LGTM, however there are no unit tests for cache and I think we should add it to be able to reach 100% coverage.

chanchiwai-ray:
Added unittests with 100% coverage for `cache.py`

PR #104, opened 2023-03-21
fix(juju_spell/config.py): Make uuid as requried

jneo8:
Before it will not raise error is uuid is not provide in config.

Reference: https://github.com/beetbox/confuse/issues/107


esunar (approved):

chanchiwai-ray (approved):
LGTM, but we should add simple test case for this.

jneo8:
> LGTM, but we should add simple test case for this.

 Added

